PR #2, opened 2024-04-29
SPICE-0001: Import Graph Analyzer API

bioball:
This adds a SPICE for an API that analyzes the imports declared by Pkl modules.

### DIFF
--- a/spices/SPICE-0001-import-graph-analyzer-api.adoc
+++ b/spices/SPICE-0001-import-graph-analyzer-api.adoc
@@ -195,7 +195,7 @@ class ImportAnalyzer {
 }
 ----
 
-The `ImportAnalyzer`  is the in-language representation of the `pkl anaylze imports` CLI command.
+The `ImportAnalyzer` is the in-language representation of the `pkl analyze imports` CLI command.
 
 === Java API
 
@@ -295,7 +295,7 @@ The `transitiveModules` property is removed.
 
 == Future directions
 
-=== `pkl:anaylze` standard library module
+=== `pkl:analyze` standard library module
 
 This is a starting point for the `pkl.analyze` stdlib module.
 Other forms of static analysis can be added to it as the need grows.
@@ -305,7 +305,7 @@ Other forms of static analysis can be added to it as the need grows.
 This feature introduces `PklAnalyzer`, a new Java API that can analyze the imports of modules without requiring the evaluator.
 This is a capability that isn't being added to other language bindings.
 
-A simple workaround is to use the standard evaluator API, given the introduction of the new `pkl.anaylze` stdlib module, albeit with the drawback of the overhead of performing Pkl evaluation.
+A simple workaround is to use the standard evaluator API, given the introduction of the new `pkl.analyze` stdlib module, albeit with the drawback of the overhead of performing Pkl evaluation.
 
 To support evaluator-free analysis, a future improvement can be to add messages to the server API for performing import graph analysis.
 
--- a/spices/SPICE-0001-import-graph-analyzer-api.adoc
+++ b/spices/SPICE-0001-import-graph-analyzer-api.adoc
@@ -1,0 +1,336 @@
+= Import Graph Anaylzer API
+
+* Proposal: xref:SPICE-0001-import-graph-analyzer-api.adoc[SPICE-0001]
+* Author: link:https://github.com/bioball[Dan Chao]
+* Status: TBD
+* Implemented in: TBD
+* Category: Tooling
+
+== Introduction
+
+Provide a way to produce a graph of imports required by Pkl modules.
+
+== Motivation
+
+Many build tools require that all inputs are known to them ahead of time, for cacheing purposes.
+
+For example, Gradle's `Task` API provides controls for file-based inputs. It uses these inputs to perform an _up-to-date check_. If the input files have not changed, it skips doing the actual work described by the task.
+
+In Pkl parlance, the inputs for any evaluation includes the dependent Pkl modules that are imported in the entrypoint program.
+Given the following program:
+
+.myFile.pkl
+[source,pkl]
+----
+import "otherFile.pkl"
+
+result = otherFile
+----
+
+The inputs for evaluation of `myFile.pkl` includes file `otherFile.pkl`.
+
+Currently, our link:https://pkl-lang.org/main/current/pkl-gradle[Gradle Plugin] requires _manual_ changes by developers to correctly configure their up-to-date checks.
+If their program adds another file for import, developers need to update the `transitiveModules` property to include this new file:
+
+.build.gradle.kts
+[source,diff]
+----
+ pkl {
+   evaluators {
+     register("evalPkl") {
+       sourceModules.set(files("myFile.pkl"))
++      transitiveModules.set(files("otherFile.pkl"))
+     }
+   }
+ }
+----
+
+This labor is tedious, and error-prone.
+If `otherFile.pkl` also has its own imports, those files need to be declared to the plugin.
+
+This type of up-to-date checking is common in other build tools. Examples:
+
+* Bazel
+* Make
+
+== Proposed Solution
+
+To address these pain points, a new API will be introduced to produce the set of imports required by an entrypoint Pkl module.
+
+The API is provided in multiple ways:
+
+1. As a CLI command: `pkl analyze imports`.
+2. As a Java API: `ImportAnalyzer`.
+3. As a Pkl standard library API: `pkl:analyze`.
+
+== Detailed Design
+
+=== What is an import?
+
+An import is either an amends/extends clause, an import clause/expression, or the result of a globbed import clause/expression.
+
+Globbed imports are not a single import, but rather, zero to many imports.
+These glob patterns are expanded into the resulting import list.
+
+Example:
+
+[source,pkl]
+----
+amends "OtherModule.pkl" // <1>
+
+import "path/to/somethingElse.pkl" // <2>
+import* "path/to/other/*.pkl" as otherModules // <3>
+
+res1 = import("some/other/module.pkl") // <4>
+
+res2 = import*("some/other/*.pkl") // <5>
+----
+<1> Amends clause
+<2> Import clause
+<3> Globbed import clause
+<4> Import expression
+<5> Globbed import expression
+
+=== CLI Command: `pkl analyze imports`
+
+A new command will be added to the CLI, called `pkl analyze imports`.
+
+Syntax:
+
+[source]
+----
+pkl analyze imports [-f,--format <format>] [--project-dir <path>] <moduleUri> [...moduleUri]
+----
+
+This command parses the provided source modules, and resolves all imported modules to their absolute URI.
+
+By default, the result is Pcf. Example:
+
+[source,shell]
+----
+pkl analyze imports myFile.pkl
+----
+
+Produces:
+[source]
+----
+imports {
+  ["file:///path/to/myFile.pkl"] {
+    "file:///path/to/otherFile.pkl"
+  }
+  ["file:///path/to/otherFile.pkl"] {}
+}
+resolvedUris {
+  ["file:///path/to/myFile.pkl"] = "file:///path/to/myFile.pkl"
+  ["file:///path/to/myFile.pkl"] = "file:///path/to/myFile.pkl"
+}
+----
+
+The value of the `imports` property is an object whose keys are module URIs, and whose values are the set of modules imported by the module referenced by the key.
+
+The set of all input modules in a program is the set of keys in the `imports` object.
+
+The value of `resolvedUris` is the mapping of in-language module URIs to those that Pkl uses to determine how to fetch the module's contents.
+For example, in the case of a link:https://pkl-lang.org/main/current/language-reference/index.html#local-dependencies[local dependency], an in-language URI might look like `projectpackage://example.com/foo@1.0.0#/mod.pkl`, and its resolved URI might look like `\file:///path/to/foo/mod.pkl`.
+
+The resolved URIs help build systems understand how resources map
+
+Underneath the hood, the command builds a synthetic module, that ultimately uses standard library module <<stdlib-module-pkl-analyze,pkl.analyze>> to produce output.
+
+==== `--format`, `-f`
+
+The `--format` (`-f` for short) flag sets the output format for this command.
+
+Examples:
+
+[source,shellscript]
+----
+# Evaluate as JSON
+$ pkl analyze imports -f json myFile.pkl
+
+# Evaluate as YAML
+$ pkl analyze imports -f yaml myFile.pkl
+----
+
+==== Common CLI flags
+
+This command recursively traverses the import graph.
+To support this, the set of flags from common evaluator settings are available:
+
+* `--allowed-modules`
+* `--allowed-resources`
+* `--cache-dir`
+* `--module-path`
+* `--project-dir`
+* `--ca-certificates`
+
+[[stdlib-module-pkl-analyze]]
+=== Standard library module `pkl:analyze`
+
+A new standard library module is introduced.
+It contains the class `ImportAnalyzer`, which provides import graph analysis.
+
+[source,pkl]
+----
+/// Utilities for static analysis of Pkl modules.
+///
+/// Unlike `pkl:reflect`, these methods do not require loading the modules for analysis.
+@ModuleInfo { minPklVersion = "0.26.0" }
+@Since { version = "0.26.0" }
+module pkl.analyze
+
+class ImportAnalyzer {
+  /// The modules to analyze for imports.
+  moduleUris: Listing<String>
+
+  /// The graph of imports declared by [moduleUris].
+  /// Each key-value pair is a module, corresponding with the set of modules
+  /// imported by that module.
+  ///
+  /// The set of all modules in the graph can be obtained via its keys.
+  external imports: Mapping<Uri, Listing<Uri>(isDistinct)>
+
+  /// Mappings of modules from their in-language URI, to the URI that is used to fetch the module's contents.
+  external resolvedUris: Mapping<Uri, Uri>(keys == imports.keys)
+}
+----
+
+The `ImportAnalyzer`  is the in-language representation of the `pkl anaylze imports` CLI command.
+
+=== Java API
+
+A Java-only API is introduced to perform actual analysis.
+
+Unlike the CLI, this API only parses a module, and does not involve the evaluator.
+
+[source,java]
+----
+public interface ImportAnalyzer {
+    Results analyze(List<ModuleSource> moduleSources);
+
+    interface Results {
+        Map<URI, Set<URI>> getImports();
+
+        Map<URI, URI> getResolvedUris();
+    }
+}
+
+----
+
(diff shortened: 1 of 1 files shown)


KushalP:
I think there are some additional use cases that could be considered as part of the implementation here.

### Use case: SBOMs

If you squint slightly when reading this SPICE, it should be possible to build an SBOM of all packages and files imported, and their version numbers from the point of view of an entrypoint. That would be useful to query out a known-bad package should you (as a user of Pkl, and not a machine or build tool) want to understand the required set of packages/imports.

### Use case: visual representation

This is a nice-to-have. It would be nice to have a documented example of how to build something like a graphviz view of the dependency graph from the produced output. It would help to understand the graph in a different way than just for build tool consumption.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+----
+import "otherFile.pkl"
+
+result = otherFile
+----
+
+The inputs for evaluation of `myFile.pkl` includes file `otherFile.pkl`.
+
+Currently, our link:https://pkl-lang.org/main/current/pkl-gradle[Gradle Plugin] requires _manual_ changes by developers to correctly configure their up-to-date checks.
+If their program adds another file for import, developers need to update the `transitiveModules` property to include this new file:
```
translatenix, line 33: I'd assume that `fileTree("baseDir") { include("**/*.pkl") }` is often good enough.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+  ///
+  /// The set of all modules in the graph can be obtained via its keys.
+  external imports: Mapping<Uri, Listing<Uri>(isDistinct)>
+
+  /// Mappings of modules from their in-language URI, to the URI that is used to fetch the module's contents.
+  external resolvedUris: Mapping<Uri, Uri>(keys == imports.keys)
+}
+----
+
+The `ImportAnalyzer`  is the in-language representation of the `pkl anaylze imports` CLI command.
```
translatenix, line 198: /
```suggestion
The `ImportAnalyzer` is the in-language representation of the `pkl analyze imports` CLI command.
```

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+[source,pkl]
+----
+/// Utilities for static analysis of Pkl modules.
+///
+/// Unlike `pkl:reflect`, these methods do not require loading the modules for analysis.
+@ModuleInfo { minPklVersion = "0.26.0" }
+@Since { version = "0.26.0" }
+module pkl.analyze
+
+class ImportAnalyzer {
```
translatenix, line 182: Why a class instead of a top-level function? `ImportAnalyzer` is a function in disguise.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+  /// The modules to analyze for imports.
+  moduleUris: Listing<String>
+
+  /// The graph of imports declared by [moduleUris].
+  /// Each key-value pair is a module, corresponding with the set of modules
+  /// imported by that module.
+  ///
+  /// The set of all modules in the graph can be obtained via its keys.
+  external imports: Mapping<Uri, Listing<Uri>(isDistinct)>
```
translatenix, line 191: Wouldn't `Map` be more efficient/capable?

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+=== Java API
+
+A Java-only API is introduced to perform actual analysis.
+
+Unlike the CLI, this API only parses a module, and does not involve the evaluator.
+
+[source,java]
+----
+public interface ImportAnalyzer {
+    Results analyze(List<ModuleSource> moduleSources);
```
translatenix, line 209: I feel the Pkl API should look more like this.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+}
+----
+<1> dependency notation turns into `projectpackage://`.
+<2> local dependency's resolve URI is a local file.
+<3> module URIs that don't resolve to anything further stay the same.
+<4> package URIs stay the same.
+
+=== Gradle Plugin Changes
+
+The input files for various Pkl Gradle tasks will switch to using the Java `ImportAnalyzer` to build its input files.
```
translatenix, line 289: How exactly will this work? Do you plan to run `ImportAnalyzer` at configuration time?

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+The set of input files will be `resolvedUris` whose keys start with `\file:///`.
+All other modules are assumed to be external to the machine and cache-able.
+
+The `transitiveModules` property is removed.
+
+== Future directions
+
+=== `pkl:anaylze` standard library module
```
translatenix, line 298:
```suggestion
=== `pkl:analyze` standard library module
```

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+This is a starting point for the `pkl.analyze` stdlib module.
+Other forms of static analysis can be added to it as the need grows.
+
+=== Support in other language bindings
+
+This feature introduces `PklAnalyzer`, a new Java API that can analyze the imports of modules without requiring the evaluator.
+This is a capability that isn't being added to other language bindings.
+
+A simple workaround is to use the standard evaluator API, given the introduction of the new `pkl.anaylze` stdlib module, albeit with the drawback of the overhead of performing Pkl evaluation.
```
translatenix, line 308:
```suggestion
A simple workaround is to use the standard evaluator API, given the introduction of the new `pkl.analyze` stdlib module, albeit with the drawback of the overhead of performing Pkl evaluation.
```

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: That's fine for simple use-cases, but it's problematic in large enough repos, especially if a single module's output doesn't need to be re-computed if an unrelated Pkl file changes.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
++  imports: Listing<Module>
+ }
+----
+
+The benefit of this approach is that an import can be defined in terms of regular Pkl evaluation.
+We do not need to provide a new CLI command, nor extension points in Gradle or server mode.
+
+Some drawbacks:
+
+* Reflection currently only provides information about a module's schema, using metadata that is already gathered when a module is initialized. Exposing this level of import requires that _data_ is also gathered during module initialization. This adds overhead to basic evaluation, even if the import graph is not needed.
```
translatenix, line 335: Import clauses/expressions are cheap to gather during parsing. What other data is required?

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+* Using `pkl:reflect` requires that Pkl initializes a module, which is much more expensive than just parsing it. Some modules are known to take seconds to initialize.
```
translatenix, line 336: What exactly is module initialization, and why can it take seconds?

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: Yeah, perhaps. I went with `Mapping` because this is designed to produce rendered output.

This:

```
imports {
  ["file:///foo.pkl"] {
     "file:///bar1.pkl"
     "file:///bar2.pkl"
     "file:///bar3.pkl"
  }
}
```

Instead of:

```
imports = Map("file:///foo.pkl", List("file:///bar1.pkl", "file:///bar2.pkl", "file:///bar3.pkl"))
```

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+=== `pkl:anaylze` standard library module
+
+This is a starting point for the `pkl.analyze` stdlib module.
+Other forms of static analysis can be added to it as the need grows.
+
+=== Support in other language bindings
+
+This feature introduces `PklAnalyzer`, a new Java API that can analyze the imports of modules without requiring the evaluator.
+This is a capability that isn't being added to other language bindings.
```
translatenix, line 306: In my opinion, the longer-term goal should be to make every feature available to language bindings and to have no privileged Java APIs.

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: This feels a little nicer:

```
analysis = new analyze.ImportAnalyzer {
  moduleUris {
    "file:///foo.pkl"
    "file:///bar.pkl"
  }
}
```

This feels a little more stunted:

```
analysis = analyze.getImports(List("file:///foo.pkl", "file:///bar.pkl"))
```

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: Module initialization is the first thing that happens when you do evaluation. The module's class is initialized, and then Pkl will execute something within the module (e.g. read `output.text`).

When initializing a class, Pkl will load all class properties, which is eager. When Pkl loads this following module, it also executes imports because they are required to initialize the types:

```
import "Bar.pkl"

bar: Bar
```

Overall, this can be expensive in large enough modules.

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: Definitely agreed! But, I am setting that as out of scope for this SPICE; this can be a future improvement.

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: I wouldn't call this "cheap"; adding runtime metadata is a new cost that we wouldn't have had to account for previously.

spices/SPICE-0001-import-graph-analyzer-api.adoc
translatenix: In my opinion, this isn’t a good use of a class. It reminds me of languages that have no choice but to simulate functions with objects. The only thing that looks odd about the function is that it doesn’t use varargs. Can’t stdlib modules use varargs?

spices/SPICE-0001-import-graph-analyzer-api.adoc
translatenix: Perhaps the rendering could use a converter?

bioball:
@KushalP:

> If you squint slightly when reading this SPICE, it should be possible to build an SBOM of all packages and files imported, and their version numbers from the point of view of an entrypoint. That would be useful to query out a known-bad package should you (as a user of Pkl, and not a machine or build tool) want to understand the required set of packages/imports.

Yeah, that's a valid use-case that can be derived from this feature. But, I would say that building SBOMs is out-of-scope for this feature itself.

> This is a nice-to-have. It would be nice to have a documented example of how to build something like a graphviz view of the dependency graph from the produced output. It would help to understand the graph in a different way than just for build tool consumption.

Yeah; good idea. `--format=dot` might be a good addition to the output formats.

spices/SPICE-0001-import-graph-analyzer-api.adoc
translatenix: I don’t know what the answer is, but this feels important to get right.

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: Sorry, I edited my previous post. See the newest paragraph for a blurb about varargs.

spices/SPICE-0001-import-graph-analyzer-api.adoc
translatenix: Does initialization include parsing? And what does loading a property entail?

spices/SPICE-0001-import-graph-analyzer-api.adoc
translatenix: The import clauses/expressions need to be parsed in any case. My assumption is that the only eager work remaining is to collect the parsed import URIs in a list.

spices/SPICE-0001-import-graph-analyzer-api.adoc
translatenix: I see. I still feel that a function is the right choice here, even if it doesn't look as pretty.

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: Fair!

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: The implementation is already in progress; we'll see.

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: The cost here is storing metadata, rather than the runtime cost of parsing.

translatenix: More details would be required to understand the given justification. In general, pkl:reflect seems like a natural place for this feature.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+  }
+  ["file:///path/to/foo.pkl"] {}
+  ["file:///path/to/my/foo1.pkl"] {}
+  ["file:///path/to/my/foo2.pkl"] {}
+  ["file:///path/to/my/subdir/foo.pkl"] {}
+}
+resolvedUris {
+  ["projectpackage://example.com/bar@1.0.0#/Bar.pkl"] = "file:///path/to/local/bar/Bar.pkl" // <2>
+  ["file:///foo/bar/Bar.pkl"] = "file:///foo/bar/bar.pkl" // <3>
+  ["package://example.com/bar@1.0.0#/Bar.pkl"] = "package://example.com/bar@1.0.0#/Bar.pkl" // <4>
```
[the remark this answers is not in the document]
bioball: I thought about that, but I can't find a URI that represents "a file within a zip file downloaded from HTTPS".

Java has jar-URIs, e.g. `jar:https://something.com/foo.zip!myFile.pkl`, but this is a Java-ism and not something generally used by other tools. Also, these URIs start with "jar" and not "zip", which is especially tied to Java.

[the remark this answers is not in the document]
stackoverflow: Yeah, you are right that it would be misleading to generate a URI that wouldn't point to the file in question. Leaving it as `package` would signal that you need to understand this scheme to properly fetch the module.

translatenix:
@stackoverflow Ideally, Java bindings would work the same as all other language bindings, and Java libs such as pkl-config-java and pkl-doc would exclusively use these bindings. This architecture shift has enormous potential.

stackoverflow:
@translatenix We already discussed creating a traditional Java binding, like the Go binding. This will also help people who can't move to newer JDKs.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+A new standard library module is introduced.
+It contains the class `ImportAnalyzer`, which provides import graph analysis.
+
+[source,pkl]
+----
+/// Utilities for static analysis of Pkl modules.
+///
+/// Unlike `pkl:reflect`, these methods do not require loading the modules for analysis.
+@ModuleInfo { minPklVersion = "0.26.0" }
+@Since { version = "0.26.0" }
```
stackoverflow, line 179: I know it's just an example, but still better to change this and `minPklVersion` to `0.27.0` to be correct.

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,336 @@
...
+A Java-only API is introduced to perform actual analysis.
+
+Unlike the CLI, this API only parses a module, and does not involve the evaluator.
+
+[source,java]
+----
+public interface ImportAnalyzer {
+    Results analyze(List<ModuleSource> moduleSources);
+
+    interface Results {
```
stackoverflow, line 211: Why isn't this a `record`? Will there be multiple implementations?

stackoverflow (approved):
Looks good! Some small comments.

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: I dropped the thread here as we were busy working on other features. But, some more thoughts:

I looked into this further, and parsing is actually what takes the lions share of Pkl's time. It turns out that Pkl's parse is _quite_ slow.

However, there are some other reasons for why we'd like this here rather than in `pkl:reflect`:

* `pkl:reflect` requires that the module requires that the module be valid, which is a limitation that we don't necessarily want for this. In these cases, it's useful for an import analyzer to still be able to run. Examples:
** A module that circularly amends itself
** A module that declares invalid members (e.g. `hidden local`)
* the analyzer computes things that we probably don't want to pay for in `pkl:reflect`. For example, it expands glob imports, and also produces resolved URIs (which might involve fetching HTTP contents).

spices/SPICE-0001-import-graph-analyzer-api.adoc
bioball: Ended up going with your suggestion, with a top-level `function importGraph(Set<Uri>)` method.

spices/SPICE-0001-import-graph-analyzer-api.adoc
translatenix: I see. Fingers crossed that a hand-rolled parser will be faster than the ANTLR parser. (AFAIK, ANTLR parsers can be quite fast.)

spices/SPICE-0001-import-graph-analyzer-api.adoc
```diff
@@ -0,0 +1,371 @@
...
+}
+
+/// An import as declared inside a module.
+class Import {
+  /// The absolute (in-language) URI of the import.
+  ///
+  /// Dependency notation URIs (such as `import "@foo/bar"`) are resolved to package URIs with
+  /// scheme `projectpackage:`.
+  uri: Uri
+}
```
bioball, line 223: FYI: changed this design slightly, and introduced class `Import`.

This allows us to evolve this over time (perhaps add `sourceLocation`, `text`).

PR #3, opened 2024-04-30
SPICE-0003: Windows Safe Paths

bioball:
Implemented in https://github.com/apple/pkl/pull/489

### DIFF
--- a/spices/SPICE-0003-windows-safe-paths.adoc
+++ b/spices/SPICE-0003-windows-safe-paths.adoc
@@ -1,0 +1,137 @@
+= Windows Safe Paths
+
+* Proposal: xref:SPICE-0003-windows-safe-paths.adoc[SPICE-0003]
+* Author: link:https://github.com/bioball[Dan Chao]
+* Status: TBD
+* Implemented in: TBD
+* Category: Language
+
+== Introduction
+
+In order to support running Pkl on Windows, filepaths are encoded to be compatible on Windows operating systems.
+
+== Motivation
+
+Currently, Pkl is only supported for macOS and Linux.
+On these operating systems, the only character reserved in filepaths is `/` (the directory separator).
+
+On Windows, the characters `<`, `>`, `:`, `"`, `\`, `|`, `?`, and `*` are also reserved.
+
+In situations where filepaths are written to disk, operations to write filepaths that contain these characters will either fail or have otherwise unexpected behavior.
+
+== Proposed Solution
+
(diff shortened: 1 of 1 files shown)


spices/SPICE-0003-windows-safe-paths.adoc
```diff
@@ -0,0 +1,137 @@
...
+It can also be a confusing experience if Pkl behaved differently on different operating systems.
+2. It preserves compatiblity with current Pkl.
+
+== Detailed design
+
+=== Filename encoding
+
+Characters that aren't safe on Windows are encoded by enclosing their hexadecimal character code in parentheses.
+
+The open parenthesis literal `(` is encoded as `((`.
```
stackoverflow, line 54: I think we should also encode `)` as `))` for consistency.
Also, why are there illegal cases if they can be properly encoded? `foo(az)bar` -> `foo((az))bar`

bioball: That example shows that `foo(az)bar` is an illegal encoding, because `az` is not a hex code.

spices/SPICE-0003-windows-safe-paths.adoc
```diff
@@ -0,0 +1,137 @@
...
+=== Prior art
+
+Kotlin's Dokka encodes Windows-unsafe characters with the decimal character code, enclosed with square brackets.
+
+A class called `class `Foo:bar`` creates a file called `Foo[58]bar.html`.
+However, this encoding uses characters that are not URI-safe, resulting in more bytes when representing them in URLs (`Foo%5B58%5Dbar.html`), and subjectively, less aesthetically pleasing URIs.
+
+== Compatibility
+
+This proposal is backwards compatibility for evaluation. However, it does mean that the cache directory needs to be re-populated.
```
translatenix, line 103: As far as I know, Gradle also reads from older cache versions (maybe even copying cache entries to the latest cache version) to avoid the need to repopulate the cache. However, this may be overkill for Pkl.

bioball: Yeah, I thought about that too. I agree that this is overkill for us (Pkl is pretty young, I don't anticipate that this will be too painful for users).

spices/SPICE-0003-windows-safe-paths.adoc
bioball: Summary from offline discussion: we will not escape the ending parenthesis (we won't encode if we don't need to).

Regex also works this way; the closing `]` in a character class doesn't need to be escaped.

PR #4, opened 2024-05-01
Add SPICE-0002 on URI rewriting

holzensp:
This adds a SPICE for URI rewriting for e.g. running Pkl in air-gapped / mirroring environments.

### DIFF
--- a/spice-templates/SPICE-0000-pkl-template.adoc
+++ b/spice-templates/SPICE-0000-pkl-template.adoc
@@ -1,6 +1,6 @@
 = SPICE Name
 
-* Proposal: [SPICE-NNNN](./SPICE-NNNN-name-of-proposal.adoc)
+* Proposal: link:./NNNN-name-of-proposal.adoc[SPICE-NNNN]
 * Status: Accepted or Rejected
 * Implemented in: Pkl <version>
 * Category: Language, Standard Library, Tooling
--- a/spices/0002-uri-rewriting-for-mirroring.adoc
+++ b/spices/0002-uri-rewriting-for-mirroring.adoc
@@ -1,0 +1,319 @@
+= SPICE 2: URI Rewriting for Mirroring
+
+* Proposal: link:./0002-uri-rewriting-for-mirroring.adoc[SPICE-0002]
+* Author: link:https://github.com/holzensp[Philip Hölzenspies]
+* Status: Review
+* Implemented in: Pkl 0.26.0
+* Category: Tooling
+
+== Introduction
+
+This SPICE discusses rewrite rules for URIs of modules being `import`ed, so as to use mirrored alternatives.
+
+== Motivation
+
+For systems that require run-time air-gapping, packages imported from `pkg.pkl-lang.org` or other sites on the internet are not available.
+This can be remedied by pre-populating a cache directory with all required dependencies (see link:../spices/0001-import-graph-analyzer.pkl[SPICE 1: Import Graph Analyzer API]).
+When dependencies can not be determined ahead of time, pre-populating the cache is not an option.
+Mirroring packages from internet repositories in a local store is insufficient, because the mirrored packages may have hard-coded dependencies.
+Importing mirrored packages, using URIs pointing to the mirror, will eventually evaluate `import`s of those hard-coded dependencies, referring to external sites.
+
+Another reason to want to consume (transitive) dependencies from a mirror is the requirement of altered versions of packages.
+For example, redirecting a package developed for `github.com` to point to a GitHub Enterprise instance.
+Note that, because upstream dependencies specify the SHA256 hash, altering packages on a mirror requires _all upstream dependants to be updated_.
+This SPICE does _not_ provide further facilities for altering packages when mirroring.
+
+== Proposed Solution
+
+Similar to `git-config`'s `insteadOf`, the proposed solution is to rewrite URIs when used in `import`, `import*`, `amends`, `extends`, `read`, and `read*` clauses/expressions.
+From `git-config`'s `man` page, on `url.<base>.insteadOf`:
+
+> Any URL that starts with this value will be rewritten to start, instead, with `<base>`.
+> In cases where some site serves a large number of repositories, and serves them with multiple access methods, and some users need to use different access methods, this feature allows people to specify any of the equivalent URLs and have Git automatically rewrite the URL to the best alternative for the particular user, even for a never-before-seen repository on the site.
+> When more than one insteadOf strings match a given URL, the longest match is used.
+>
+> Note that any protocol restrictions will be applied to the rewritten URL.
+> If the rewrite changes the URL to use a custom protocol or remote helper, you may need to adjust the `protocol.*.allow` config to permit the request.
+
+The solution proposed in this SPICE mirrors this design, but extends it to allow regular expressions for the match.
+Only after rewriting are URIs tested for compliance with Allowed Modules and Allowed Resources settings.
+
+Rewrites can be provided:
+
+1. Via the API;
+2. Via `pkl:Project`; and
+3. Via arguments on the CLI.
+
+== Detailed design
+
+=== API
+
+The API changes described here refer to the Java API, but this SPICE implies analogous changes to the Swift and Go APIs.
+
+The `EvaluatorBuilder` gets the following new methods:
+
+[source,java]
+----
+public final class EvaluatorBuilder {
+    // ...
+
+    /**
+     * Adds the given URI rewrite rule.
+     *
+     * @param replace   The literal string to match.
+     * @param with      The literal string to substitute for the first match of [replace].
+     */
+    public EvaluatorBuilder addRewriteRule(String replace, String with) {
+        // ...
+    }
+
+    /**
+     * Adds the given URI rewrite rule.
+     *
+     * @param replace   The regular expression to match.
+     * @param with      The string to substitute for the first match of [replace], which may contain
+     *                  references to capture groups (`$0`, `$1`, etc).
+     */
+    public EvaluatorBuilder addRewriteRule(Pattern replace, String with) {
+        // ...
+    }
+}
+----
+
+=== Standard library module `pkl:Project`
+
+Rewrite rules can be defined project-wide by declaring them in the `PklProject` file.
+
+[source,pkl]
+----
+@Since { version = "0.26.0" }
+class UriRewrite {
+  /// What to match in the given URI.
+  replace: String|Regex
+
+  /// What to replace the _all_ matches with.
+  ///
+  /// When `replace` is a `String`, `with` is interpreted literally.
+  /// When `replace` is a `Regex`, `with` is interpreted as a replacement pattern,
+  /// which may contain references to capture groups (`$0`, `$1`, etc).
+  with: String
+}
+
+/// Rewrite rules for URIs used in module keys and resource reads.
+///
+/// For any URI, the first of the longest-matching [UriRewrite]s is applied, before applying the secutiry manager.
+@Since { version = "0.26.0" }
+uriRewrites: Listing<UriRewrite>
+----
+
+=== Common CLI flags
+
+These flags are _common_, because this setting applies to different commands equally.
+The proposed `analyze` command (see link:../spices/0001-import-graph-analyzer.pkl[SPICE 1: Import Graph Analyzer API]), for example, applies these rewrites in the resolution of all URIs analyzed and reported.
+
+This proposal adds two flags to the CLI:
+  * `--uri-rewrite <regex> <replacement>`, which adds a URI rewrite rule; and
+  * `--no-uri-rewrites`, which discards all URI rewrite rules defined thus far.
+
+If a `pkl:Project` file is found, this is read first.
+To invoke Pkl without rewrite rules defined therein, `--no-uri-rewrites` can be used, followed by zero-or-more `--uri-rewrite` arguments.
+
+=== Rewrite behavior
+
+The proposed design is to match any URI being resolved in its entirety to all the provided rewrites' `replace` specification.
+The first, longest-matching rewrite rule is applied.
+In other words:
+ - when multiple rules match, a rule that matches a longer substring of the URI is preferred over one matching a shorter substring; and
+ - when multiple rules match the same substring length, the first rule is chosen (in definition order).
+
+The intended rewrite behavior, expressed as Pkl, is
+[source, pkl]
+----
+class Rewrite {
+  rewrites: List<UriRewrite>
+  uri: Uri
+
+  matchesByScore = rewrites.groupBy((rewrite) ->
+    if (!uri.contains(rewrite.replace))
+      -1
+    else if (rewrite.replace is Regex)
+      let (match = rewrite.replace.findMatchesIn(uri).first)
+        match.end - match.start
+    else
+      rewrite.replace.length
+  )
+
+  maxScore = matchesByScore.keys.max
+
+  rewritten: Uri = if (maxScore == -1) uri else
+    let (rewrite = matchesByScore[maxScore].first)
+      uri.replaceFirst(rewrite.replace, rewrite.with)
+}
+
+function rewrite(inputUri: Uri, uriRewrites: List<UriRewrite>): Uri = new Rewrite {
+  rewrites = uriRewrites
+  uri = inputUri
+}.rewritten
+----
+
+=== Implementation
+
+The proposed design is to implement this behaviour in `IoUtils::resolve`.
+`IoUtils::resolve` is used throughout to resolve URIs, regardless of whether a URI is used in an `import`, `import*`, `extends`, `amends`, `read`, or `read*`.
+
+=== Error messages
+
+Rewriting URIs according to definitions _not_ in the modules being evaluated could lead to hard to debug failures.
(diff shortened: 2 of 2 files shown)


spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /**
+     * Adds the given URI rewrite rule.
+     *
+     * @param replace   The regular expression to match.
+     * @param with      The string to substitute for the first match of [replace], which may contain
+     *                  references to capture groups (`$0`, `$1`, etc).
+     */
+    public EvaluatorBuilder addRewriteRule(Pattern replace, String with) {
```
bioball, line 77: To be consistent with allowed modules, allowed resources, I would just have `Pattern` here.

Might be good to have a `record RewriteRule(Pattern replace, String with)`; and to have:

```
addRewriteRule(RewriteRule rewriteRule)

setRewriteRules(Collection<RewriteRule> rewriteRules)
```

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+=== Standard library module `pkl:Project`
+
+Rewrite rules can be defined project-wide by declaring them in the `PklProject` file.
+
+[source,pkl]
+----
+@Since { version = "0.26.0" }
+class UriRewrite {
+  /// What to match in the given URI.
+  replace: String|Regex
```
bioball, line 92: I think using `pkl.Regex` as a configuration value is an anti-pattern, because it is specifically a Java 8 compatible regex, and not portable across languages.

Following that, I think this should be a `String` instead:

```
/// A regular expression pattern describing what to match in a given URI.
///
/// Regular expressions follow Java rules.
@SourceCode { language = "RegExp" }
replace: String(isRegex)
```

And `with` unconditionally accepts placeholders for capture groups.

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+If a `pkl:Project` file is found, this is read first.
+To invoke Pkl without rewrite rules defined therein, `--no-uri-rewrites` can be used, followed by zero-or-more `--uri-rewrite` arguments.
+
+=== Rewrite behavior
+
+The proposed design is to match any URI being resolved in its entirety to all the provided rewrites' `replace` specification.
+The first, longest-matching rewrite rule is applied.
+In other words:
+ - when multiple rules match, a rule that matches a longer substring of the URI is preferred over one matching a shorter substring; and
+ - when multiple rules match the same substring length, the first rule is chosen (in definition order).
```
bioball, line 127: Since these rewrite rules are user-defined as an in-order list, I think we can just pick the first rewrite rule that matches.

It would also be cheaper this way; it allows us to short-circuit once we find a rule that matches.

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+at attempt#dependency (file:///Users/jappleseed/demo/attempt.pkl, line 1)
+
+3 | name = dependency.name
+           ^^^^^^^^^^
+at attempt#name (file:///Users/jappleseed/demo/attempt.pkl, line 3)
+
+106 | text = renderer.renderDocument(value)
+             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/0.25.2/stdlib/base.pkl#L106)
+----
```
bioball, line 225: How would this work if the rewrite rule is provided via another means, like via a CLI flag?

Also, it doesn't sound great to insert stack frames to point to nodes that aren't actually being executed. We do this sparingly (e.g. to represent values that initialized as parse-time constants, and a language implementation optimization detail).

Rather than this, I suggest that we decorate the error with hints. Take a look at `VmExceptionBuilder#withHint`.

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+
+When URI rewrite rules are used to ensure no remote resources are used at all, a rule to check "all URIs not touched by any of the rewrite rules are rejected" can be expressed.
+This rule is based on the fact that it matches at zero-width, so no other rule has a lower score:
+
+[source,pkl]
+----
+  new UriRewrite {
+    replace = Regex("^")
+    with = throw("URI found that is uncovered by any of the rewrite rules")
+  }
```
bioball, line 318: Not sure if this can work--in order to configure the evaluator at runtime, all of the rewrite rules need to be consumed (ergo, deep forced).

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
+= SPICE 2: URI Rewriting for Mirroring
+
+* Proposal: link:./0002-uri-rewriting-for-mirroring.adoc[SPICE-0002]
+* Author: link:https://github.com/holzensp[Philip Hölzenspies]
+* Status: Review
+* Implemented in: Pkl 0.26.0
```
bioball, line 6:
```suggestion
* Status: TBD
* Implemented in: TBD
```

spice-templates/SPICE-0000-pkl-template.adoc
```diff
@@ -1,6 +1,6 @@
 = SPICE Name
 
-* Proposal: [SPICE-NNNN](./SPICE-NNNN-name-of-proposal.adoc)
+* Proposal: link:./NNNN-name-of-proposal.adoc[SPICE-NNNN]
```
bioball, line 3: Nit: this is causing a merge conflict; rebase and resolve

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+> If the rewrite changes the URL to use a custom protocol or remote helper, you may need to adjust the `protocol.*.allow` config to permit the request.
+
+The solution proposed in this SPICE mirrors this design, but extends it to allow regular expressions for the match.
+Only after rewriting are URIs tested for compliance with Allowed Modules and Allowed Resources settings.
+
+Rewrites can be provided:
+
+1. Via the API;
+2. Via `pkl:Project`; and
+3. Via arguments on the CLI.
```
bioball, line 45: There's also:
* Gradle plugin
* Java evaluator API
* Server mode message passing API
* `~/.pkl/settings.pkl` file

In particular I think the settings file is an important place to be able to configure rewrite rules. Not everything needs to be a project, and in some cases can't be a project. This enables regular evaluation of Pkl without needing to introduce projects.

It also is really useful for, say, if you have a machine image (e.g. Docker) that you want to distribute that is already configured to talk to an internal host.

spices/0002-uri-rewriting-for-mirroring.adoc
bioball: We might also want to validate that a rewrite can only be rewritten to the same scheme, and perhaps that the rewrite is only for http/https.

Rewriting to a different scheme can be problematic because of minor differences between how different schemes work (e.g. HTTPS urls don't support glob imports).

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+function rewrite(inputUri: Uri, uriRewrites: List<UriRewrite>): Uri = new Rewrite {
+  rewrites = uriRewrites
+  uri = inputUri
+}.rewritten
+----
+
+=== Implementation
+
+The proposed design is to implement this behaviour in `IoUtils::resolve`.
+`IoUtils::resolve` is used throughout to resolve URIs, regardless of whether a URI is used in an `import`, `import*`, `extends`, `amends`, `read`, or `read*`.
```
bioball, line 162: This should probably use `org.pkl.core.http.RequestRewritingClient` instead, because `IoUtils.resolve` doesn't affect how packages get loaded.

We can definitely use `org.pkl.core.http.RequestRewritingClient` if we only permit `http`/`https` rewrites. It'd be more problematic if we allow other types of rewrites though.

spices/0002-uri-rewriting-for-mirroring.adoc
holzensp: I was hoping for an easy way to allow non-regex pattern specification, to not have to always have to handle escaping (in the URI case, especially of `.`) and I wasn't considering the case where Pkl's own configuration would be object-mapped to a host language.

Mind you, `Regex` uses `Pattern.compile(pattern, Pattern.UNICODE_CHARACTER_CLASS | Pattern.UNICODE_CASE)`, where `isRegex` uses `Pattern.compile(self, Pattern.UNICODE_CASE)`, so they're the same Java language level, regardless, and the former is slightly more expressive.

spices/0002-uri-rewriting-for-mirroring.adoc
holzensp: So we want this to intentionally exclude custom resource readers?

spices/0002-uri-rewriting-for-mirroring.adoc
holzensp: "Via the API" was intended to refer to the Java Evaluator API (hence the proposed changes to `EvaluatorBuilder`). If this is reworded, are you expecting support for other APIs?

Similarly, how do we want to do these for SPICEs generally? List all surfaces where some work needs doing? Then it also needs to list every supported language binding / build system. I was generally assuming it understood that any change/addition of the Evaluator API _implies_ changes/additions to build systems / server / etc. Basically; do we want SPICEs to show the explicit design for _every_ contact surface?

spices/0002-uri-rewriting-for-mirroring.adoc
holzensp: ^^ This goes back to whether we want to restrict rewriting to HTTP(S) and ignore custom resource readers.

spices/0002-uri-rewriting-for-mirroring.adoc
bioball: Yeah, since we are dealing with Java regexes here, it's quite nice that we can use the `isRegex` helper.

The pattern that I was saying we should follow is to use strings to represent regex values, which is a universal way to describe them (`String` is a suitable type no matter the regex flavor).

IMO: `Regex` should only be used for in-language string matching.

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+
+== Motivation
+
+For systems that require run-time air-gapping, packages imported from `pkg.pkl-lang.org` or other sites on the internet are not available.
+This can be remedied by pre-populating a cache directory with all required dependencies (see link:../spices/0001-import-graph-analyzer.pkl[SPICE 1: Import Graph Analyzer API]).
+When dependencies can not be determined ahead of time, pre-populating the cache is not an option.
+Mirroring packages from internet repositories in a local store is insufficient, because the mirrored packages may have hard-coded dependencies.
+Importing mirrored packages, using URIs pointing to the mirror, will eventually evaluate `import`s of those hard-coded dependencies, referring to external sites.
+
+Another reason to want to consume (transitive) dependencies from a mirror is the requirement of altered versions of packages.
```
translatenix, line 21: This amounts to monkey patching. Is it wise for a deterministic config language to encourage/support this?

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+
+== Introduction
+
+This SPICE discusses rewrite rules for URIs of modules being `import`ed, so as to use mirrored alternatives.
+
+== Motivation
+
+For systems that require run-time air-gapping, packages imported from `pkg.pkl-lang.org` or other sites on the internet are not available.
+This can be remedied by pre-populating a cache directory with all required dependencies (see link:../spices/0001-import-graph-analyzer.pkl[SPICE 1: Import Graph Analyzer API]).
+When dependencies can not be determined ahead of time, pre-populating the cache is not an option.
```
translatenix, line 17: How can this happen?

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+public final class EvaluatorBuilder {
+    // ...
+
+    /**
+     * Adds the given URI rewrite rule.
+     *
+     * @param replace   The literal string to match.
+     * @param with      The literal string to substitute for the first match of [replace].
+     */
+    public EvaluatorBuilder addRewriteRule(String replace, String with) {
```
translatenix, line 66: I think the method name should be more specific. Suggestions: `addModuleUriRewriteRule`, `addImportUriRewriteRule’, ‘rewriteImportUri`. When I see `String replace`, I’m thinking that’s the replacement.

spices/0002-uri-rewriting-for-mirroring.adoc
translatenix: Users need to understand the exact regex semantics. “follow Java rules” is too vague, and `isRegex` only covers regex syntax. Because the regex will be evaluated by Pkl (even if it happens in Java code), `replace: Regex` seems logical to me. I don’t see why portability matters here.

spices/0002-uri-rewriting-for-mirroring.adoc
```diff
@@ -0,0 +1,319 @@
...
+
+* Proposal: link:./0002-uri-rewriting-for-mirroring.adoc[SPICE-0002]
+* Author: link:https://github.com/holzensp[Philip Hölzenspies]
+* Status: Review
+* Implemented in: Pkl 0.26.0
+* Category: Tooling
+
+== Introduction
+
+This SPICE discusses rewrite rules for URIs of modules being `import`ed, so as to use mirrored alternatives.
```
translatenix, line 11: This is a scary feature. I’m worried that it might cause more harm than good. Does Pkl do “feature previews” like (say) Java? Does this feature exist in other languages, especially in languages with a similar safety mindset?

spices/0002-uri-rewriting-for-mirroring.adoc
translatenix: I think all places that are expected to require changes should at least be mentioned, especially those in apple/pkl. Supporting a new feature in the Evaluator API doesn’t necessarily mean that it should also be supported in the Gradle plugin.

spices/0002-uri-rewriting-for-mirroring.adoc
bioball: I don't think this is a valid use-case; packages have their integrity verified by checksum. Altering a package would break any integrity checks that we would do.

spices/0002-uri-rewriting-for-mirroring.adoc
bioball: > Users need to understand the exact regex semantics. “follow Java rules” is too vague (and why refer Pkl users to Java

I disagree with this, actually. When writing regex, it's much easier to understand "this is a Java regex" than needing to look through the full set of available. It's not any different than the many tools that refer their users to Perl (see [PCRE](https://en.wikipedia.org/wiki/Perl_Compatible_Regular_Expressions)).

As a matter of principle, `Regex` shouldn't be used to express configuration data in Pkl; it's a data type that's meant for in-language operations. The same can be said of `IntSeq`, `Pair`, `List`, `Map`, etc.

spices/0002-uri-rewriting-for-mirroring.adoc
bioball: This is naturally an opt-in feature. Pkl would not do any rewriting unless specifically configured to do so. Unless you were thinking of something else? E.g. a library author being able to opt out of rewrites?

spices/0002-uri-rewriting-for-mirroring.adoc
HT154: Doing this in `HttpClient` seems at odds with support for other schemes, doesn't it? At the very least it would make for a confusing or undiscoverable API.

spices/0002-uri-rewriting-for-mirroring.adoc
HT154: Would it be potentially possible to supporting rewriting between schemes that share the same hierarchical/globbable(/local?) properties?

HT154 (commented):
How/when would rewrites interact with `package://` and `projectpackage://` reads? Would rewrite rules apply before resolution from `projectpackage` to `package`? When determining what underlying `https` URIs are needed for a particular `package`? To the `https` URIs for the package manifest and zip? Multiple? All?

Should rewrite rules always apply to both module and resource reads? What about cross-scheme rewrite cases where a module reader is registered with a scheme but no corresponding resource reader exists? Maybe it makes sense to have each rewrite rule specify which reader type(s) it applies to? Or alternatively define separate sets of rules? eg.

```pkl
uriRewrites: Listing<UriRewrite>

moduleUriRewrites: Listing<UriRewrite> = uriRewrites

resourceUriRewrites: Listing<UriRewrite> = uriRewrites
```

Largely a not: While the language style guide doesn't appear to specify, I'm personally in favor of retaining abbreviation capitalization in APIs, so `uriRewrite: Listing<URIRewrite>`, but I'm happy to defer to a style on this if it takes a stance.

PR #5, opened 2024-05-02
SPICE-0004: HTTP Proxy Support

bioball:
This is a continuation of the design discussion started here: https://github.com/apple/pkl/discussions/309

Implementation: https://github.com/apple/pkl/pull/472

### DIFF
--- a/spices/SPICE-0004-HTTP-Proxy-Support.adoc
+++ b/spices/SPICE-0004-HTTP-Proxy-Support.adoc
@@ -1,0 +1,180 @@
+// suppress inspection "HttpUrlsUsage" for whole file
+= HTTP Proxy Support
+
+* Proposal: link:./SPICE-0004-HTTP-Proxy-Support.adoc[SPICE-0004]
+* Author: link:https://github.com/bioball[Dan Chao]
+* Status: TBD
+* Implemented in: TBD
+* Category: Language, Standard Library
+
+== Introduction
+
+This SPICE describes support for HTTP proxying.
+
+== Motivation
+
+Some environments do not allow direct connections to the internet, instead requiring that connections go through an intermediary proxy.
+
+To enable Pkl in these situations, this SPICE provides a way to redirect outbound requests through said proxies.
+
+== Proposed Solution
+
+Pkl will support HTTP proxies.
+Proxy settings will be added, and this will control how Pkl makes HTTP requests.
+
+== Detailed design
+
+=== Configuring proxies
+
+A proxy can be set via the following, in order of precedence (least to greatest):
+
+1. `~/.pkl/settings` file
+2. `PklProject` file (in `evaluatorSettings`)
+3. CLI Flags/pkl-gradle options/`HttpClient#Builder`/server mode message passing API
+
+==== Environment Variables
+
+Proxies will _not_ be configured via environment variables.
+This maintains consistency with the rest of Pkl; it does not support using environment variables for any other configuration.
+Additionally, while other languages might use environment variables, they have slightly different meaning from one language to another (see link:https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy[]]).
+Not using environment variables avoids a possble problem where the same environment variable influences both Pkl and another runtime, but in different ways.
+
+==== Settings file
+
+A new `proxy` property is added to the global settings file.
+
+[source,pkl]
+----
+module pkl.settings
+
+import "pkl:Proxy"
+
+/// HTTP proxy settings.
+@Since { version = "0.26.0" }
+proxy: Proxy?
+----
+
+==== `PklProject`
+
+A new `proxy` property is added to `EvaluatorSettings` of a PklProject file.
+
+[source,pkl]
+----
+module pkl.Project
+
+import "pkl:Proxy"
+
+class EvaluatorSettings {
+  proxy: Proxy?
+}
+----
+
+[[new-stdlib-module]]
+==== New stdlib module: `pkl.Proxy`
+
+A new stdlib module is added called `pkl.Proxy`, that is used by `pkl.settings` and `pkl.Project`.
+
+[source,pkl]
+----
+/// Settings that control how Pkl talks to HTTP proxies.
+@Since { version = "0.26.0" }
+module pkl.Proxy
+
+/// The proxy to use for HTTP(s) connections.
+///
+/// At the moment, only HTTP proxies are supported.
+///
+/// Example:
+/// ```
+/// local proxyPassword = read("my-proxy-password.txt").text
+/// address = "http://myuser@\(proxyPassword):my.proxy.example.com:5080"
+/// ```
+address: Uri(startsWith("http://"))
+
+/// Hosts to disable connecting to via a proxy.
+///
+/// Values can be either hostnames, or CIDR blocks.
+///
+/// Follows the same rules as cURL's [--noproxy](https://curl.se/docs/manpage.html#--noproxy)
+/// flag.
+noProxy: Listing<String>(isDistinct)
+----
+
+==== CLI flags
+
+The following CLI flags are added to common CLI options:
+
+* `--proxy` -- A proxy URL to connect to
+* `--noproxy` -- A comma-separated list of hosts to not proxy.
+
+==== pkl-gradle
+
+The following options are added to the common options for `PklSpec`:
+
+* `proxy: Property<String>`
+* `noProxy: ListProperty<String>`
+
+==== `HttpClient#Builder`
+
+When Pkl is embedded with in Java, it is preferrable to use the host's proxy settings if possible.
+When Pkl is run standalone, the proxy should be configued via user-facing means.
+
+To support this, `HttpClient#Builder` receives two new builder methods:
+
+* `setProxySelector(java.net.ProxySelector proxySelector)`
+* `setProxy(URI proxyAddress, List<URI> noProxy)`
+
+The default builder receives `java.util.ProxySelector#getDefault`, which uses host proxy settings if one is set.
+This is the proxy selector that gets used by `EvaluatorBuilder#preconfigured`.
+
+`setProxy(URI proxyAddress, List<URI> noProxy)` sets a proxy selector that follows the noProxy rules described in <<noproxy>>,
+and is what the CLI, Gradle, and message passing API use.
+
+=== Proxy Configuration Values
+
+There are two values to configure for proxying:
+
+==== Address
+
+The proxy address specifies the proxy host, protocol, and optional port.
+It takes the form of a standard URL. For example:
+
+[source,text]
+----
+"http://my.proxy.com:8080"
+----
+
+The scheme _must_ be `http`.
+
+[[noproxy]]
+==== `noProxy`
+
+The noProxy value follows the specification outlined in link:https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/#standardizing-no_proxy[standardizing `noproxy`], except in how it is configured.
+
+1. The `*` wildcard matches all hosts.
+2. A hostname with a leading dot (`.`) is stripped.
+3. A hostname matches all domain prefixes.
+4. CIDR blocks can be specified for matching.
+5. Hostnames are not resolved; a hostname will match its declared name only and not its possibly resolved IP address.
+6. IP addresses are treated verbatim (no special handling of loopback address).
+
+== Compatibility
+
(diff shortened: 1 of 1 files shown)


spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+A proxy can be set via the following, in order of precedence (least to greatest):
+
+1. `~/.pkl/settings` file
+2. `PklProject` file (in `evaluatorSettings`)
+3. CLI Flags/pkl-gradle options/`HttpClient#Builder`/server mode message passing API
+
+==== Environment Variables
+
+Proxies will _not_ be configured via environment variables.
```
translatenix, line 37: I think this is a mistake. Env vars are *the* way to configure proxy settings for command-line tools. Unlike the listed hypothetical worries, having to configure each tool individually is a very real problem.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+module pkl.settings
+
+import "pkl:Proxy"
+
+/// HTTP proxy settings.
+@Since { version = "0.26.0" }
+proxy: Proxy?
+----
+
+==== `PklProject`
```
translatenix, line 57: What’s the use case for this?

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+----
+
+==== CLI flags
+
+The following CLI flags are added to common CLI options:
+
+* `--proxy` -- A proxy URL to connect to
+* `--noproxy` -- A comma-separated list of hosts to not proxy.
+
+==== pkl-gradle
```
translatenix, line 110: What’s the use case case for having different proxy settings for Pkl than for Gradle?

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+* `noProxy: ListProperty<String>`
+
+==== `HttpClient#Builder`
+
+When Pkl is embedded with in Java, it is preferrable to use the host's proxy settings if possible.
+When Pkl is run standalone, the proxy should be configued via user-facing means.
+
+To support this, `HttpClient#Builder` receives two new builder methods:
+
+* `setProxySelector(java.net.ProxySelector proxySelector)`
```
translatenix, line 124: Is it really necessary to expose ProxySelector?

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+import "pkl:Proxy"
+
+class EvaluatorSettings {
+  proxy: Proxy?
+}
+----
+
+[[new-stdlib-module]]
+==== New stdlib module: `pkl.Proxy`
```
translatenix, line 73: Having a stdlib module just for proxy settings feels strange.
Can’t this be part of pkl.settings, even if pkl.Project needs it too (which I’m not convinced of)?

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+When Pkl is embedded with in Java, it is preferrable to use the host's proxy settings if possible.
+When Pkl is run standalone, the proxy should be configued via user-facing means.
+
+To support this, `HttpClient#Builder` receives two new builder methods:
+
+* `setProxySelector(java.net.ProxySelector proxySelector)`
+* `setProxy(URI proxyAddress, List<URI> noProxy)`
+
+The default builder receives `java.util.ProxySelector#getDefault`, which uses host proxy settings if one is set.
```
translatenix, line 127:
```suggestion
The default builder receives `java.net.ProxySelector#getDefault`, which uses host proxy settings if one is set.
```

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+Java applications that embed Pkl and construct an evaluator with `EvaluatorBuilder#preconfigured` will be configured to talk to the host's proxy settings, if one is set via `java.net.ProxySelector.setDefault()`.
+This is technically a breaking change.
+
+== Future directions
+
+There are several more features related to proxies that can be supported in the future:
+
+* HTTPS proxies
+* SOCKS proxies
+* Alternative authentication methods (e.g. setting the `Proxy-Authorization` header)
```
translatenix, line 172: Alternative to what? As far as I can tell, this spice doesn’t propose any auth support (which I believe is the right way to start given how little is offered by the JDK).

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: I'm open to the argument, but: another concern here is: how does this play with `--env-var`? What happens if you do:

```
pkl eval --env-var http_proxy=http://some.proxy foo.pkl
```

It feels like this shouldn't affect proxy settings, because this is ostensibly just a way to configure what `read("env:")` will return.

But if that doesn't affect the proxy, this is possibly a strange behavior:

```
# talk via the proxy
export http_proxy=http://some.proxy pkl eval foo.pkl

# does not talk via the proxy
pkl eval --env-var http_proxy=http://some.proxy foo.pkl
```

There's also plenty of languages/tools whose proxies aren't configured by env vars. Just some quick skimming (as far as I can tell):

* Node.js
* Java
* Rust, Cargo
* PHP

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: I am deploying into an environment that needs proxying. Instead of setting it up on my target machine, or setting it up via a script, I can just set it in my PklProject file.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: I want to control `pkl eval`  through Gradle, but I want it to otherwise behave exactly like my CLI does.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: I'm open to ideas here!

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: Originally, I figured we would just use basic auth, but I'm backing off on that idea due to security concerns.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: Both `pkl.settings` and `pkl.Project` need this, and neither is a logical subset of the other, so it feels cleanest to put it in a separate module.

We can also put it in `base.pkl`, but that affects the global namespace, which doesn't feel great.

We also possibly should put redirect rules in there, so maybe this intermediary module should be called `HttpSettings`, with proxy and rewrite rules both being in there.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
translatenix: The correct proxy settings depend on the network you’re on. I can’t think of any use case where it is useful to have different proxy settings for Gradle and the Gradle Pkl plugin.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
translatenix: Why did you expose it?

spices/SPICE-0004-HTTP-Proxy-Support.adoc
translatenix: Do you have evidence that this is useful? Is it practical to only have per-project proxy settings?

spices/SPICE-0004-HTTP-Proxy-Support.adoc
translatenix: Java supports OS proxy settings, and there are signs that env vars will be reconsidered. Supporting neither OS settings nor env vars seems problematic, forcing users to script a solution specifically for Pkl.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
translatenix: If a new stdlib module is needed, its scope should definitely be wider than proxy settings.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: The goal is to allow users to configure Pkl's proxy the same way their application's proxy is configured. `setProxy(proxyAddress, List<URI> noProxy)` is our bespoke implementation, and accepting a `java.net.ProxySelector` allows those that are already using it to configure this client the same way.

Some prior art; both `java.net.HttpClient` and okhttp accept `ProxySelector`.

translatenix: Of course java.net.HttpClient accepts ProxySelector. The point is that with every java.net type added, it becomes more difficult to ever have another implementation of org.pkl.util.http.HttpClient (say NettyHttpClient). It’s better to wait until there is a proven need for “those that are already using ProxySelector can configure our client the same way”, then make an informed decision. That’s why I didn’t expose HttpClient.sendAsync—it’s not yet needed, and with virtual threads added in Java 21, it may never be needed.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
stackoverflow: Not sure I'm sold on configuring proxies via env vars. Nothing else in Pkl in configured using them, also, It may lead to "works on my machine" problems where you have the right env vars set.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
stackoverflow: I think the rationale here is that `settings.pkl` is a user specific configuration, which only leaves `PklProject` and CLI/Evaluator options left for shared configs. Configuring your proxy every time via CLI doesn't seem ideal, so adding this settings to `PklProject` allows for a project-wide proxy configuration.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+class EvaluatorSettings {
+  proxy: Proxy?
+}
+----
+
+[[new-stdlib-module]]
+==== New stdlib module: `pkl.Proxy`
+
+A new stdlib module is added called `pkl.Proxy`, that is used by `pkl.settings` and `pkl.Project`.
```
stackoverflow, line 75: Calling it `pkl.HttpSettings` would future-proof if for when URL rewriting arrives, though this can be done in said SPICE.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+There are several more features related to proxies that can be supported in the future:
+
+* HTTPS proxies
+* SOCKS proxies
+* Alternative authentication methods (e.g. setting the `Proxy-Authorization` header)
+
+== Alternatives considered
+
+TBD
```
stackoverflow, line 176: URL rewriting was a considered alternative, but after some thought we found out they don't fully overlap and there may be a need for both.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: > But with every java.net type added, it becomes more difficult to ever have another implementation of org.pkl.util.http.HttpClient (say NettyHttpClient).

I'm not sure why this is desirable. Can you elaborate?

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: Supporting OS proxy settings sounds more enticing than supporting envvar-based configuration, if that's possible.

spices/SPICE-0004-HTTP-Proxy-Support.adoc
bioball: > Do you have evidence that this is useful? Is it practical to only have per-project proxy settings?

Yeah, this is something that was brought up as a desired use-case from some of our users.

PR #7, opened 2024-05-20
SPICE-0006: Import Paths on Windows

bioball:
Implementation within https://github.com/apple/pkl/pull/492

### DIFF
--- a/spices/SPICE-0006-file-paths-on-windows.adoc
+++ b/spices/SPICE-0006-file-paths-on-windows.adoc
@@ -1,0 +1,144 @@
+= File paths on Windows
+
+* Proposal: link:./SPICE-0006-file-paths-on-windows.adoc[SPICE-0006]
+* Status: TBD
+* Implemented in: TBD
+* Category: Language
+
+== Introduction
+
+On Windows, relative paths will use the `/` character as the filepath separator.
+
+== Detailed Design
+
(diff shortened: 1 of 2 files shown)


spices/SPICE-0006-file-paths-on-windows.adoc
```diff
@@ -0,0 +1,144 @@
...
++fileContents = read("file:///\(uri.encode(pwdAdjusted))/foo.json")
+----
+
+== Future directions
+
+=== Current working directory API
+
+To improve scripts that <<reading-off-pwd,read off `PWD`>>, Pkl can provide an API to provide the current working directory that is platform independent.
+
+This might require a new standard library module.
```
holzensp, line 122: `PWD` is just one (typical) environment variable. There are many `FOO_HOME`-style environment variables that would be similarly impacted. Is an alternative, maybe, to have a `normalizedWindowsPath` property (or a much better name) on `String` a more general alternative?

holzensp (approved):

spices/SPICE-0006-file-paths-on-windows.adoc
bioball: Yeah; another idea is to have a path API, where you have path constructors, expose the current working dir as a path object.

stackoverflow (approved):
Agree on only accepting `/` as path separator. We should provide some library to deal with paths and OSes, but that's another SPICE.

PR #8, opened 2024-06-05
SPICE-0007: Const checks in typealiases

bioball:
Implementation in https://github.com/apple/pkl/pull/516


stackoverflow (approved):
Agree that having typealiases being late-bound is not worth the complexity it brings to the semantics of the language. It also makes them in line with classes.

PR #9, opened 2024-07-27
Add SPICE for `delete`

holzensp:


spices/SPICE-0008-member-deletion.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+[source,pkl]
+----
+amends "base.pkl"
+
+prop {
+  [0] = super[0] - 14
+  496
+  ["quux"] = foo
+  ["baz"] = delete // <1>
```
stackoverflow, line 133: Have you consider an alternate syntax for `delete`?
This seems very similar to normal assigning syntax and, specially without syntax highlighting, may fly under the radar while reviewing code.

What about

```pkl
prop {
  delete ["entry"] // for entries
  delete 0 // for indexes
  delete prop // for properties
  delete [[foo == 10]] // for predicates
}
```

It's easier to see the change and totally unambiguous for the parser.

holzensp: I had considered this, but opted against it for a few reasons (that I'll include in `Alternatives considered`); I like that in your change, deletes are emphasised over other amends/overwrites. The downsides, however:
* currently and with my proposed syntax, you can skim the left-margin of an amends expression to find all things that are affected
* your changes requires member predicates and entries to occur in different places (namely in the argument position of `delete`) than they do now
* `delete prop`, to me, reads like an expression that would be evaluated as usual, i.e. resolving and evaluating `prop` to use its _value_ rather than its _name_, which is particularly painful considering your index suggestion (without square brackets); surely that could be done parametrically ;)
* as shown here, such alternative syntax makes room for all sorts of small variations (here, that element indices don't have square brackets) that allow subtle bugs (we've been talking about ways to disambiguate entry-indices from integer-valued entry keys), and require more syntax knowledge of the user.

bioball (commented):
Overall, this approach makes sense to me.

Some general concerns about deletion:

* Shifting object elements seems quite dangerous. The element index is also the member's name. This means that deletion isn't just a delete, it's also a rename of existing members. This can produce quite confusing errors and make Pkl code harder to reason through.
* As another feature in the future, we want to improve provenance of object members. For example, we want to show in error messages where an object value was originally defined. If we shift object elements, it probably makes this task more complicated; we would need to convey "this object defined `[1] = 4`, and then a deletion happened so it changed to `[0] = 4`"
* I'm curious how the implementation should work here. This changes our object model quite a bit--right now, the evaluation of an object works by collecting all member names from the root-most object in the prototype chain, then evaluating all member values from the top-most object down (see `VmObject.forceAndIterateMemberValues()`). If our members can change their names (as in the case of element deletion), our model of evaluation will need to change. Also: how does this affect performance?

As far as I can tell, none of the other config languages offer a first-class delete mechanism, so there is not much prior art to draw from.

spices/SPICE-0008-member-deletion.adoc
MForever78: I actually like what @stackoverflow 's proposal better.

> * currently and with my proposed syntax, you can skim the left-margin of an amends expression to find all things that are affected

Is this just for readability, or has any impact to the implementation? This feels like a nice to have feature. But `= delete` feels odd semantically. Most people would think the right side of `=` to be a value instead of an operator.

> * `delete prop`, to me, reads like an expression that would be evaluated as usual, i.e. resolving and evaluating `prop` to use its _value_ rather than its _name_, which is particularly painful considering your index suggestion (without square brackets); surely that could be done parametrically ;)

Would `delete [0]` address the concern? It feels reasonable to add brackets around indices.

translatenix:
Hiding instead of deleting members would keep references intact while still removing members from output. I think it should be considered as an alternative to deletion.

holzensp:
The problem with hiding members is that the indices become non-contiguous. This means that, for example, `myListing.toList().last == myListing[myListing.length - 1]` doesn't reliably hold anymore. Similarly, `for (i, v in myListing) { when (i > 0) { f(myListing[i - 1], v) } }` breaks. In other words; it only moves the problem. See the `Preserving indices of elements` section.

translatenix:
> The problem with hiding members is that the indices become non-contiguous.

I don’t see why. By “hiding” I mean the same behavior as `hidden`—nothing changes when accessing the member from Pkl code (hence no holes and no dangling references), but the member is skipped by renderers.

bioball:
The issue that I see with "hiding" these members is that you get _different_ names when you then `toList()`:

```pkl
l1 = new Listing { 1; 2; 3 } { [0] = delete } // new Listing { (hidden) 1; 2; 3 }
l2 = l1.toList() // List(2, 3)

res1 = l1[0] // 1
res2 = l2[0] // 2
```

And also that causes havoc with equals:

```pkl
l1 = new Listing { 1; 2; 3 } { [0] = delete }
l2 = new Listing { 2; 3 }

res = l1 == l2 // is this true or false?
res2 = l1[0] == l2[0] // if the above is true, shouldn't this be true?
```

Going one further:

```pkl
l1 = new Listing { 1; 2; 3 } { [0] = delete }
l2 = l1.toList().toListing()

res = l1 == l2 // shouldn't this be true?
```

translatenix:
@bioball If `[0] = hidden` only affects output rendering, none of the problems you mentioned exists. (I’m not claiming this solution is perfect–it’s all about tradeoffs.) I believe Jsonnet supports hiding for objects fields.

holzensp:
```
todos: Listing<String>

enumeratedTodos: String = new Listing {
  for (number, item in todos) {
    "\(number). \(item)"
  }
}
```
Question; would this enumeration of TODOs have non-contiguous numbers, or would hidden things show up here?

translatenix:
> Question; would this enumeration of TODOs have non-contiguous numbers, or would hidden things show up here?

I’d imagine that iterating over hidden elements/entries would behave the same as iterating over hidden object members does today.

PR #10, opened 2024-08-06
SPICE-0009: External Readers

HT154:
Implemented in:
* https://github.com/apple/pkl/pull/660
* https://github.com/apple/pkl-go/pull/101
* https://github.com/apple/pkl-swift/pull/26

### DIFF
--- a/spices/SPICE-0009-external-readers.adoc
+++ b/spices/SPICE-0009-external-readers.adoc
@@ -72,39 +72,156 @@ If a codebase or user wishes to enforce that only secure LDAP connections are ma
 
 To avoid terminology confusion with the existing language binding message passing model, the `pkl` process will remain known as the "server" and the reader process will be known as the "client", despite all requests in this proposal being initiated by the `pkl` process.
 
-* Message Passing API
-** Add DiscoverReadersRequest
-** Add DiscoverReadersResponse
-** CreateEvaluatorRequest
-*** `externalReaders`
-** Pull into a new pkl-msgapi(?) package shared by pkl-code and pkl-server
-*** Message*.kt files at least
+=== Overall Flow
 
-* Java API
-** New `ExternalModuleKeyFactory` type
-*** Entrypoints (CLI, Server) construct based on `externalReaders`
-*** On (initialization? explicit startup?), launch subprocess
-*** send each DiscoverReadersRequest, await DiscoverReadersResponse (timeout? 1s?)
+1. User requests an evaluator (via Java API, Message Passing API, or CLI) with an external reader.
+2. Entrypoint (client Java code, `CliEvaluator`, or `Server`) instantiates an `ExternalReader`.
+3. The `ExternalReader` is started, which starts the configured child process and sends a `DiscoverReadersRequest` message.
+4. A `DiscoverReadersResponse` message is received (within some timeout).
+5. The entrypoint requests resource readers (`ExternalResourceReader`) and module key factories (`ExternalModuleKeyFactory`) from the `ExternalReader` and uses them to build an evaluator.
+`ExternalResourceReader` and `ExternalModuleKeyFactory` function similarly to the existing `ClientResourceReader` and `ClientModuleKeyFactory`, implementation may be reusable entirely.
+6. Evaluation begins.
+7. A resource or module read is encountered.
+8. The appropriate `Read*Request` message is sent to the child process, the corresponding `Read*Response` message is awaited.
+9. Evaluation continues.
+10. Evaluation completes.
+11. The `ExternalReader` is closed, terminating the child process.
 
-* stdlib
-** `pkl:EvaluatorSettings` (used in `pkl:Project`)
-*** `externalReaders: Listing<ExternalReader>`
+=== Java API
 
-* CLI
-** `--external-reader`
-*** use `clikt.multiple`
-*** may be passed as a space-separated string where the first element becomes `executable` and any remainder becomes `arguments`
-*** should this be shlex'd? or just split on spaces?
+New APIs:
 
-* Language Bindings
-** Add `ExternalReaderRuntime`
-** Add `externalReaders` to `EvaluatorOptions`
+* `ExternalResourceReader` - a `ResourceReader` implementation similar (identical?) to `ClientResourceReader`.
+* `ExternalModuleKeyFactory` - a `ModuleKeyFactory` implementation similar (identical?) to `ClientModuleKeyFactory`.
+* `ExternalModuleKey` - a `ModuleKey` implementation similar (identical?) to `ClientModuleKey`.
+* `ExternalReader` - manages the lifecycle of child processes.
+    ** Explicit `open`/`close` methods to manage child process lifecycle.
+    ** Opening an `ExternalReader` spawns the subprocess, sets up the `MessageTransport`, sends a `DiscoverReadersRequest` message, and awaits a `DiscoverReadersResponse` response.
+    ** Methods to procure `ExternalResourceReader` and `ExternalModuleKeyFactory` instances.
+
+This proposal requires that the message passing API functionality move out of pkl-server.
+A new project pkl-msgapi will be created to contain the new APIs and the core messaging code current part of pkl-server (`pkl-server/src.main/kotlin/org.pkl.server/Message*.kt`).
+The pkl-msgapi project will be depended on by pkl-cli, pkl-server, and Java API clients using external readers.
+
+=== Message Passing API
+
+Two new message types will be added:
+
+[source,pkl]
+----
+/// Code: 0x2e
+/// Type: Server Request
+class DiscoverReadersRequest {
+  /// A number identifying this request
+  requestId: Int
+}
+
+/// Code: 0x2f
+/// Type: Client Response
+class DiscoverReadersResponse {
+  /// A number identifying this request
+  requestId: Int
+
+  /// Register client-side module readers.
+  ///
+  /// [ClientModuleReader] is defined at https://pkl-lang.org/main/current/bindings-specification/message-passing-api.html#create-evaluator-request
+  clientModuleReaders: Listing<ClientModuleReader>?
+
+  /// Register client-side resource readers.
+  ///
+  /// [ClientResourceReader] is defined at https://pkl-lang.org/main/current/bindings-specification/message-passing-api.html#create-evaluator-request
+  clientResourceReaders: Listing<ClientResourceReader>?
+}
+----
+
+`CreateEvaluatorRequest` will be expanded with an additional property:
+[source,pkl]
+----
+externalReaders: Listing<ExternalReader>?
+
+class ExternalReader {
+  /// May be specified as an absolute path to an executable
+  /// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
+  executable: String
+
+  /// Command line arguments that will be passed to the reader process
+  arguments: Listing<String>
+}
+----
+
+=== CLI
+
+A new `--external-reader` CLI argument will be added to configure external readers.
+This argument can be provided multiple times (using `clikt.multiple`) to configure multiple external readers.
+The argument may be passed as a space-separated string where the first element becomes `executable` and any remainder becomes `arguments`.
+
+TBD: It might be best if the argument value is link:https://docs.python.org/3/library/shlex.html#shlex.split[shlex'd] instead of split to support passing arguments to the reader process that contain spaces.
+
+=== Standard Library
+
+The `EvaluatorSettings` module will be expanded to enable configuring external readers in `PklProject` files:
+
+[source,pkl]
+----
+externalReaders: Listing<ExternalReader>?
+
+class ExternalReader {
+  /// May be specified as an absolute path to an executable
+  /// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
+  executable: String
+
+  /// Command line arguments that will be passed to the reader process
+  arguments: Listing<String>
+}
+----
+
+=== Language Binding Libraries
+
+The language binding libraries `pkl-go` and `pkl-swift` will be expanded to support using and implementing external readers.
+For the purpose of illustration, examples will be provided using Golang.
+
+The `EvaluatorOptions` type will be expanded to include a new property for external readers:
+
+[source,go]
+----
+type EvaluatorOptions struct {
+	// ...
+    ExternalReaders []ExternalReader
+	// ...
+}
+
+type ExternalReader struct {
+	Executable string
+	Arguments []string
+}
+----
+
+A new `ExternalReaderRuntime` type will be introduced to implement the child process message passing interface.
+It makes sense to expand the existing libraries to add this functionality as much of the message passing infrastructure and types for implementing resource and module readers can be reused.
+An `ExternalReaderRuntime` is configured with zero or more `ResourceReader` instances and zero or more `ModuleReader`.
+When started, the runtime will consume messages from standard input, dispatch calls to the configured readers, and send responses over standard output.
+
+[source,go]
+----
+type ExternalReaderRuntime interface {
+  Run()
+  Close()
+}
+
+func NewExternalReaderRuntime(resourceReaders []ResourceReader, moduleReaders []ModuleReader) ExternalReaderRuntime {
+  // ...
+}
+----
 
 == Compatibility
 
-This proposal is purely additive.
+From a language perspective, this proposal is purely additive.
 
-In the case where newer language bindings configure external readers against an older `pkl` binary, the new `CreateEvaluatorRequest.externalReaders` field will be ignored silently. If module evaluation relies on configured external readers, it will fail accordingly.
+In the case where newer language bindings configure external readers against an older `pkl` binary, the new `CreateEvaluatorRequest.externalReaders` field will be ignored silently.
+If module evaluation relies on configured external readers, it will fail accordingly.
+
+Any usage of the pkl-server APIs that are moving to pkl-msgapi will break.
+It's unlikely there are clients of these APIs outside the apple/pkl repo.
 
 == Future directions
 
@@ -115,6 +232,8 @@ This would also make it very convenient to provide friendly, type-safe Pkl APIs 
 +
 [source,pkl]
 ----
+import "pkl:json"
+
 typealias LDAPResult = Mapping<String, Listing<String>>
 
 class LDAPQuery {
@@ -152,6 +271,10 @@ email = queryResults[0]["mail"][0]
 ** Change could be made as followup work with only changes to Java APIs and internals
 
 == Alternatives considered
+
+=== Message Passing code in pkl-core
+
+Instead of creating a new project for pkl-msgapi, move the message passing code from pkl-server directly into pkl-core.
 
 === One shot, per-read subprocesses
 
--- a/spices/SPICE-0009-external-readers.adoc
+++ b/spices/SPICE-0009-external-readers.adoc
@@ -1,0 +1,120 @@
+= External Readers
+
+* Proposal: link:./SPICE-0009-external-readers.adoc[SPICE-0009]
+* Status: TBD
+* Implemented in: Pkl 0.27 ()
+* Category: Language
+
+== Introduction
+
+This SPICE proposes a way for custom module and resource readers to be used when evaluating modules with `pkl eval`.
+
+== Motivation
+
+When used via language binding libraries, custom module and resource readers schemes may be implemented by client code.
+Custom readers help bridge Pkl into systems where it does't fit natively, enabling a wide variety of use cases:
+
+* Mediating access to resources that require authentication (eg. a secret store)
+* Querying data accessible via non-HTTP protocols (eg. LDAP)
+* Providing filesystem abstractions for remote or in-memory modules
+* And more!
+
+One of the primary drawbacks to custom readers is that using one instantly turns a Pkl module into a _incompatible language_: the module may only be evaoluated by the client tool or application providing the reader and can no longer be evaluated directly using `pkl eval`.
+This hinders development and debugging of such modules, as typical workflows involving `trace(<expr>)` or `pkl eval -x '<expr>'` behave differently or are no longer possible.
+
+== Proposed Solution
+
+The existing message passing API and client libraries will be expanded to support external readers.
+Instead of client libraries invoking `pkl server`, `pkl eval` invocations, when configured with an external reader, will run the reader executable as a subprocess.
+External readers can provide any number of resource and/or module schemes and will configured as an executable and list of arguments via:
+
+* CLI flag
+* PklProject property
+* Java API
+* Message passing API
+
+
+[source,pkl]
+----
+/// May be specified as an URI to a readable resource, eg. `file:`, `https:`, or `package:`
+/// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
+executable: String
+
+/// Command line arguments that will be passed to the reader process
+arguments: Listing<String>
+----
+
+=== Example
+
+Consider this module:
+
+[source,pkl]
+----
+username = "john_appleseed"
+
+email = read("ldap://ds.example.com:389/dc=example,dc=com?mail?sub?(uid=\(username))").text
+----
+
+Pkl doesn't implement the `ldap:` resource URI scheme natively, but an external reader can provide it.
+Assuming a hypothetical `pkl-ldap` binary implementing the external reader protocol and the `ldap:` scheme is in the `$PATH`, this module can be evaluated as:
+
+[source,text]
+----
+$ pkl eval <module> --external-reader=pkl-ldap --allowed-resources prop:,ldap:
+username = "john_appleseed"
+email = "appleseed@example.com"
+----
+
+For security reasons, schemes provided by external readers are not allowed by default and clients must explicitly allow the desired scheme(s).
+In this example, the external reader may provide both `ldap:` and `ldaps:` schemes.
+If a codebase or user wishes to enforce that only secure LDAP connections are made, the `ldaps:` scheme may be allowed while `ldap:` is not.
+
+== Detailed design
+
+To avoid terminology confusion with the existing language binding message passing model, the `pkl` process will remain known as the "server" and the reader process will be known as the "client", despite all requests in this proposal being initiated by the `pkl` process.
+
+* Message Passing API
+** Add DiscoverReadersRequest
+** Add DiscoverReadersResponse
+** CreateEvaluatorRequest
+*** `externalReaders`
+** Pull into a new pkl-msgapi(?) package shared by pkl-code and pkl-server
+*** Message*.kt files at least
+
+* Java API
+** New `ExternalModuleKeyFactory` type
+*** Entrypoints (CLI, Server) construct based on `externalReaders`
+*** On (initialization? explicit startup?), launch subprocess
+*** send each DiscoverReadersRequest, await DiscoverReadersResponse (timeout? 1s?)
+
+* stdlib
+** `pkl:EvaluatorSettings` (used in `pkl:Project`)
+*** `externalReaders: Listing<ExternalReader>`
+
+* CLI
+** `--external-reader`
+*** use `clikt.multiple`
+*** may be passed as a space-separated string where the first element becomes `executable` and any remainder becomes `arguments`
+*** should this be shlex'd? or just split on spaces?
+
+* Language Bindings
+** Add `ExternalReaderRuntime`
+** Add `externalReaders` to `EvaluatorOptions`
+
+== Compatibility
+
+This proposal is purely additive.
+
+In the case where newer language bindings configure external readers against an older `pkl` binary, the new `CreateEvaluatorRequest.externalReaders` field will be ignored silently. If module evaluation relies on configured external readers, it will fail accordingly.
+
+== Future directions
+
+* Java library for bindings to support being an external reader client
+* To improve CLI ergonomics, could implement additive `--allow-resources`/`--allow-modules` args (current flags replace full list)
+* Manage external reader processes separately from EvaluatorImpl lifetime
+** Potential large savings in per-evaluator overhead for Java API and Language Binding usage
+** Savings for CLI usage (primary use case) would be minimal
+** Code is more complicated (need an ExternalReaderManager sort of mechanism tracking unique commands => processes)
+** Change could be made as followup work with only changes to Java APIs and internals
+
+== Alternatives considered
--- a/spices/SPICE-0009-external-readers.adoc
+++ b/spices/SPICE-0009-external-readers.adoc
@@ -80,22 +80,22 @@ To avoid terminology confusion with the existing language binding message passin
 
 === Overall Flow
 
-1. User requests an evaluator (via Java API, Message Passing API, or CLI) with an external reader.
-2. The entrypoint (client Java code, `CliEvaluator`, or `Server`) instantiates an `ExternalProcess` based on the configured external readers.
-3. The entrypoint instantiates resource readers (`ResourceReaders.External`) and module key factories (`ModuleKeyFactories.External`) referencing the `ExternalProcess` and uses them to build an evaluator.
+. User requests an evaluator (via Java API, Message Passing API, or CLI) with an external reader.
+. The entrypoint (client Java code, `CliEvaluator`, or `Server`) instantiates an `ExternalProcess` based on the configured external readers.
+. The entrypoint instantiates resource readers (`ResourceReaders.External`) and module key factories (`ModuleKeyFactories.External`) referencing the `ExternalProcess` and uses them to build an evaluator.
 `ResourceReaders.External` and `ModuleKeyFactories.External` function almost identically to the existing `ClientResourceReader` and `ClientModuleKeyFactory` and pkl-server will be updated to reuse these implementations where possible.
-4. Evaluation begins.
-5. A resource or module read is encountered.
-6. Upon first read only:
-1.. The `ExternalProcess` is started, which starts the configured child process.
-2.. The child process is sent a `Initialize(Module|Resource)ReaderRequest` message.
-3.. A `Initialize(Module|Resource)ReaderResponse` message is received (within some timeout).
-7. The appropriate `Read*Request` message is sent to the child process, the corresponding `Read*Response` message is awaited.
-8. Evaluation continues.
-9. Evaluation completes.
-10. The evaluator is closed.
-1.. The `ExternalProcess` is closed, the child process is sent SIGTERM.
-2.. If the child process has not terminated after some timeout (3 seconds?), it is forcefully stopped via SIGKILL.
+. Evaluation begins.
+. A resource or module read is encountered.
+. Upon first read only:
+.. The `ExternalProcess` is started, which starts the configured child process.
+.. The child process is sent a `Initialize(Module|Resource)ReaderRequest` message.
+.. A `Initialize(Module|Resource)ReaderResponse` message is received (within some timeout).
+. The appropriate `Read*Request` message is sent to the child process, the corresponding `Read*Response` message is awaited.
+. Evaluation continues.
+. Evaluation completes.
+. The evaluator is closed.
+.. The `ExternalReader` is closed, the child process is sent the `CloseExternalProcess` message.
+.. If the child process has not terminated after the set timeout (3 seconds), it is forcefully stopped via SIGKILL.
 
 === Java API
 
@@ -113,7 +113,24 @@ The code added to pkl-core will include the new APIs and the core messaging code
 
 === Message Passing API
 
-Four new message types will be added:
+`CreateEvaluatorRequest` will be expanded with additional properties:
+[source,pkl]
+----
+externalModuleReaders: Mapping<String, ExternalReader>?
+
+externalResourceReaders: Mapping<String, ExternalReader>?
+
+class ExternalReader {
+  /// May be specified as an absolute path to an executable
+  /// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
+  executable: String
+
+  /// Command line arguments that will be passed to the reader process
+  arguments: Listing<String>
+}
+----
+
+Five new message types will be added:
 
 [source,pkl]
 ----
@@ -150,7 +167,7 @@ class InitializeResourceReaderRequest {
   scheme: String
 }
 
-/// Code: 0x102
+/// Code: 0x103
 /// Type: Client Response
 class InitializeResourceReaderResponse {
   /// A number identifying this request.
@@ -162,24 +179,14 @@ class InitializeResourceReaderResponse {
   /// [ClientResourceReader] is defined at https://pkl-lang.org/main/current/bindings-specification/message-passing-api.html#create-evaluator-request
   spec: ClientResourceReader?
 }
+
+/// Code: 0x104
+/// Type: Server One Way
+class CloseExternalProcess {}
 ----
 
-`CreateEvaluatorRequest` will be expanded with additional properties:
-[source,pkl]
-----
-externalModuleReaders: Mapping<String, ExternalReader>?
-
-externalResourceReaders: Mapping<String, ExternalReader>?
-
-class ExternalReader {
-  /// May be specified as an absolute path to an executable
-  /// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
-  executable: String
-
-  /// Command line arguments that will be passed to the reader process
-  arguments: Listing<String>
-}
-----
+The `CloseExternalProcess` message exists primarily because different operating systems provide different abilities to gracefully stop child processes.
+Using an in-band message for this purpose reduces the need for external reader developers to address OS-specific implementation details.
 
 === CLI
 
--- a/spices/SPICE-0009-external-readers.adoc
+++ b/spices/SPICE-0009-external-readers.adoc
@@ -26,7 +26,7 @@ This hinders development and debugging of such modules, as typical workflows inv
 
 The existing message passing API and client libraries will be expanded to support external readers.
 Instead of client libraries invoking `pkl server`, `pkl eval` invocations, when configured with an external reader, will run the reader executable as a subprocess.
-External readers can provide any number of resource and/or module schemes and will configured as an executable and list of arguments via:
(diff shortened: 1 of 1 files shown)


spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,120 @@
...
+In this example, the external reader may provide both `ldap:` and `ldaps:` schemes.
+If a codebase or user wishes to enforce that only secure LDAP connections are made, the `ldaps:` scheme may be allowed while `ldap:` is not.
+
+== Detailed design
+
+To avoid terminology confusion with the existing language binding message passing model, the `pkl` process will remain known as the "server" and the reader process will be known as the "client", despite all requests in this proposal being initiated by the `pkl` process.
+
+* Message Passing API
+** Add DiscoverReadersRequest
+** Add DiscoverReadersResponse
```
bioball, line 78: What do these do?

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,120 @@
...
+*** `externalReaders: Listing<ExternalReader>`
+
+* CLI
+** `--external-reader`
+*** use `clikt.multiple`
+*** may be passed as a space-separated string where the first element becomes `executable` and any remainder becomes `arguments`
+*** should this be shlex'd? or just split on spaces?
+
+* Language Bindings
+** Add `ExternalReaderRuntime`
```
bioball, line 101: What is `ExternalReaderRuntime`?

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,120 @@
...
+* PklProject property
+* Java API
+* Message passing API
+
+
+[source,pkl]
+----
+/// May be specified as an URI to a readable resource, eg. `file:`, `https:`, or `package:`
+/// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
+executable: String
```
bioball, line 41: How would a URI work here? Like, how would `executable = "file:///path/to/foo"` work?

bioball (commented):
Some high level thoughts:

Using a persistent process with message passing isn't a bad idea--for processes with a higher startup cost (e.g. JVM-based processes), you pay that penalty once for the life of the evaluator, rather than every time a `read()` happens. However, there's an argument to be made for having the external reader binary be short-lived. For example, this makes it possible to write an external reader that's simply a bash script. And making these things bash-script-able is quite nice:

```bash
#!/usr/bin/env bash

set -x

URI="$1"

if [[ "$URI" != ldap:* ]]; then
  exit 1
fi

ldapsearch -D "${URI:5}"
```

If it's a persistent process, it should probably be spawned the first time a module or resource is read, then kept alive while the evaluator session is alive (e.g. until it is closed).

I don't know if we need a `DiscoverReadersRequest` and `DiscoverReadersResponse`? Wouldn't the evaluator already know about the external readers it needs?

spices/SPICE-0009-external-readers.adoc
HT154: Ah right these were in separate notes. In Pkl:
```pkl
class DiscoverReadersRequest {
  /// A number identifying this request
  requestId: Int
}

class DiscoverReadersResponse {
  /// A number identifying this request
  requestId: Int

  /// Register client-side module readers.
  clientModuleReaders: Listing<ClientModuleReader>?

  /// Register client-side resource readers.
  clientResourceReaders: Listing<ClientResourceReader>?
}
```

Where `ClientModuleReader ` and `ClientResourceReader ` use the same definitions as [`CreateEvaluatorRequest`](https://pkl-lang.org/main/current/bindings-specification/message-passing-api.html#create-evaluator-request)'s properties.

spices/SPICE-0009-external-readers.adoc
HT154: Open to suggestions on the name, but this would be the functionality in the language binding libraries that runs the subprocess-side message passing actor, eg. in pkl-go (API obviously not final)
```go
type ExternalReaderRuntime interface {
  Run()
  Close()
}

func NewExternalReaderRuntime(resourceReaders []ResourceReader, moduleReaders []ModuleReader) ExternalReaderRuntime {
  // ...
}
```

bioball: Ah, I see.

It seems a bit strange to expect clients themselves to spawn a sub-process? If we support this API in server mode, I'd expect pkl to spawn these external readers, rather than the clients, because:

* Pkl has all the information it needs already (what executable to spawn, what arguments to pass)
* Implementations can diverge between language bindings

spices/SPICE-0009-external-readers.adoc
HT154: My main hope with supporting URIs is that external reader binaries could be packaged directly in Pkl packages. A PklProject using such a reader might look something like this:
```pkl
import "pkl:platform"

dependencies {
  ["ldap"] {
    uri = "package://...@x.y.z"
  }
}

evaluatorSettings {
  allowedResources {
    "ldap:"
    "ldaps:"
  }
  externalReaders {
    new {
      executable = "@ldap/reader_\(platform.current.operatingSystem.name)_\(platform.current.processor.architecture)"
      // support for project dependency notation would be ideal!
    }
  }
}
```

Pkl would need to know how to translate `file:` URIs to absolute paths and for `https:`, `package:`, or `projectpackage:` URIs would download the file/package and execute the reader from some tmp/cache path.

Enabling this workflow would avoid complicating distribution of external reader binaries and would allow packages to provide nicer (not stringly-typed) APIs for their readers. For example, the `ldap:` example in the proposal requires detailed understanding of [RFC 4516](https://docs.ldap.com/specs/rfc4516.txt). It might be nicer if an LDAP Pkl package existed that provided an API like this:
```pkl
protocol: *"ldap"|"ldaps"
host: String
port: UInt16 = 389
baseDN: String
attributes: Listing<String>
scope: *"base"|"one"|"sub"
filter: String = "(&)" // matches anything

typealias Result = Mapping<String, Listing<String>>

fixed results: Listing<Result> = new json.Parser { useMapping = true }.parse(
  read("\(protocol)://\(host):\(port)/\(baseDN)?\(attributes.join(","))?\(scope)?\(filter)").text
)
```

stackoverflow (commented):
I'm missing an example that shows the whole process. Your example just shows a Pkl file using the reader and a cli call, but it's not clear to me how these things glue together.

Also think having the processes being long-running would put too much hurdles for people who want to contribute a reader, so let's keep that as a future improvement.

For the `package` support:
That seems to complicated the design quite a bit. Now Pkl has to understand how to properly process *packages-as-readers*. If my reader is a Ruby script and I want to deploy it as a package, do I have to bundle the script together? Gets even more complicated with executables and multiplatform support. Or maybe I misunderstood you and package readers have to be Pkl-only.
I'm not sure how useful writing readers in Pkl instead of general languages will be. Though perhaps I'm wrong. My point is, maybe pkl packages are not the best distribution mechanism for readers, even though I can't think of a good solution for this.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+email = read("ldap://ds.example.com:389/dc=example,dc=com?mail?sub?(uid=\(username))").text
+----
+
+Pkl doesn't implement the `ldap:` resource URI scheme natively, but an external reader can provide it.
+Assuming a hypothetical `pkl-ldap` binary implementing the external reader protocol and the `ldap:` scheme is in the `$PATH`, this module can be evaluated as:
+
+[source,text]
+----
+$ pkl eval <module> --external-reader=pkl-ldap --allowed-resources prop:,ldap:
```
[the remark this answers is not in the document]
HT154: In this example it is an executable, as briefly described under Proposed Solution:
```
/// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
executable: String
```

The end-to-end flow, as proposed, would be:
1. User invokes `pkl eval <module> --external-reader=pkl-ldap --allowed-resources prop:,ldap:`
2. `pkl` looks up `pkl-ldap` in `$PATH` and invokes it
3. `pkl` sends the `DiscoverReadersRequest` message
4. `pkl-ldap` responds with `DiscoverReadersResponse`. This may register zero or more resource or module reader schemes. In this case, it would make sense to register resource readers for the `ldap` and `ldaps` schemes.
5. Evaluation begins
6. Upon encountering `read("ldap:...")`, `pkl` sends a `ReadResourceRequest` (the same as it would to a language bindings client resource reader today)
7. The external reader process responds with `ReadResourceResponse`
8. Evaluation continues

spices/SPICE-0009-external-readers.adoc
HT154: I think I might not have been clear? `ExternalReaderRuntime` would be provided by language binding libraries as a means to easily implement the "client" side of external reader binaries. It would be invoked by pkl itself and not the server client.

Here's what a mostly complete entrypoint to a simple `pkl-ldap` implementation might look like in Go:
```go
func main() {
  ldapReader := ldapReader{scheme: "ldap"} // implementation elided, not important to the point
  ldapsReader := ldapReader{scheme: "ldaps"}
  runtime := pkl.NewExternalReaderRuntime([]ResourceReader{ldapReader, ldapsReader}, nil)
  runtime.Run()
}
```

Building this into the existing language binding libraries is a convenience because the message-passing infrastructure already exists there. I can't think of a situation where a pkl server client and an external reader client would exist in the same application, as it would make more sense to use a client reader in the pkl server client directly.

HT154:
Thanks for the review @bioball!

> Using a persistent process with message passing isn't a bad idea--for processes with a higher startup cost (e.g. JVM-based processes), you pay that penalty once for the life of the evaluator, rather than every time a `read()` happens. However, there's an argument to be made for having the external reader binary be short-lived. For example, this makes it possible to write an external reader that's simply a bash script.

I definitely agree that there's utility in enabling the super simple scripting route! My primary motivation for a persistent subprocess is, as you've identified, startup cost.

Here's a real world example: I would like to be able to use an external reader to mediate access to a secret store in a large monorepo. I'd estimate that there are 80+ unique resource URIs that would be read during a "world eval". The process that would implement the external reader currently takes ~0.25 seconds to invoke from the command line (in the happy path), which would add ~20 seconds to the overall evaluation time. Most of that quarter second is process and initialization overhead and the actual read takes ~0.05 seconds, so if the reader process were persistent it would save ~15 seconds from the total evaluation time.

If persistent reader processes are implemented, it also becomes fairly trivial to implement one-shot readers. Spitballing, it might look something like this:
```
pkl eval <module.pkl> --external-reader 'pkl-cmd ldap=path/to/pkl-ldap.sh ldaps=path/to/pkl-ldap.sh' --allowed-resources ...
```
Where the `pkl-cmd` executable implements the external reader message passing API and registers a ResourceReader for each specified scheme->script mapping. I'd be interested in shipping `pkl-cmd` as part of the Pkl project.

> If it's a persistent process, it should probably be spawned the first time a module or resource is read, then kept alive while the evaluator session is alive (e.g. until it is closed).
>
> I don't know if we need a `DiscoverReadersRequest` and `DiscoverReadersResponse`? Wouldn't the evaluator already know about the external readers it needs?

The ergonomic motivation for spawning a persistent process on startup and having the readers be discovered at runtime would be to decouple which reader binaries are used from which schemes they provide. If one-shot reader processes are used, the scheme->binary registration is unavoidable, but persistent readers may implement many schemes. Having to perform the same mapping for persistent processes would be very verbose (since setting allowed modules is also required), but would be able to spawn the process on first use instead of during evaluator initialization. Another consideration here is user error: if a user registers a reader for a scheme that it does not provide, how should pkl respond?

HT154:
Thanks for the review @stackoverflow!

> I'm missing an example that shows the whole process. Your example just shows a Pkl file using the reader and a cli call, but it's not clear to me how these things glue together.

I think I've addressed this in other comments, but please let me know if you still have questions.

> Also think having the processes being long-running would put too much hurdles for people who want to contribute a reader, so let's keep that as a future improvement.

My main concerns with only targeting "one-shot" reader processes at this time are:
* It's not clear how to support globbing (i.e. `ListResourcesRequest`/`ListModulesRequest`)
* It may make "persistent" readers more complicated to implement in the future from a configuration/CLI perspective
* Performance, as outlined above

Above I proposed a `pkl-cmd` persistent external reader that could function as an "adapter" of sorts to simpler one-shot reader processes. I definitely agree that the simplicity of writing a simple script should be a goal here, and I don't think the more involved persistent reader design rules that out.

> For the `package` support: That seems to complicated the design quite a bit. Now Pkl has to understand how to properly process _packages-as-readers_.

I may be misinterpreting this concern, but I'm not sure this is so complicated. Reader executables would need to be supplied as package file URIs including the fragment `#/path/to/binary`.

> If my reader is a Ruby script and I want to deploy it as a package, do I have to bundle the script together?

My thought here is to primarily target languages for which (message-passing) bindings already exist. Both Golang and Swift (at least on linux) can produce statically linked binaries that can be packaged dependency-free. If there are bindings for interpreted languages in the future, distribution would likely need to be via the language package manager, which this design does not rule out.

> Gets even more complicated with executables and multiplatform support.

Definitely agree, but I think `pkl:platform` makes this reasonable. I show what this might look like in a PklProject here: https://github.com/apple/pkl-evolution/pull/10#discussion_r1708244384

spices/SPICE-0009-external-readers.adoc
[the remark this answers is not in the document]
stackoverflow: > pkl sends the DiscoverReadersRequest message

Ah, that was the missing piece.

This would mean for every external reader, Pkl would have to call it at least twice: at startup and then again for the actual read. Not a fan of the performance problems this might bring.
I think we can go with a simpler process of configuring the reader scheme statically. Something like (just an example, the syntax can be more ergonomic):

```bash
pkl eval <module> --external-reader ldap=pkl-ldap
```
Which wouldn't require Pkl to discover schemes at runtime.

stackoverflow:
> I think I've addressed this in other comments, but please let me know if you still have questions.

You addressed it in the comments, but someone reading the SPICE shouldn't have to read comments to understand it, so it would be nice if that was clear in the SPICE itself.

stackoverflow:
About having readers being persistent:

I think you are approaching this from the wrong side. For Pkl, the reader should just be a process it spawns once it finds a read. If you want it to be persistent (which I consider an edge case) you can run a persistent background process and write a simple script for Pkl to call which would pipe the reads.
This would simplify your proposal and there wouldn't be a need for a `pkl-cmd`.

Also it's not clear to me how errors and the communication happens. I imagine we can keep the unix tradition of using return codes so 0 would mean success and other numbers would be errors. `stderr` could contain the error message in case things went wrong and `stdout` would contain the actual result of the read. But anyway, all of this should be in the SPICE.

spices/SPICE-0009-external-readers.adoc
bioball: Oooh I see--this is code for the _reader_ itself, rather than from an application calling into Pkl.

bioball:
Josh actually brings up a good point--the process of resolving a read can involve multiple calls to a reader, which can make a short-lived execution quite complicated. For example, resolving a glob star import (e.g. `import*("foo:/**.pkl")` is linear to the number of directories needed to read.

Additionally, Pkl wants to know three facts about each scheme:

* Is it globbable?
* Is it hierarchical?
* Is it local?

It's quite a rough experience to provide all this to Pkl as CLI args: this is quite long: `--external-reader ldap=pkl-ldap --globbable-reader=pkl-ldap --hierarchical-reader=pkl-ldap --local-reader=pkl-ldap`.

I echo Islon's concerns about distributing binaries using Pkl packages. I see the value here, but this isn't what packages are designed for. Besides, there are already many ways to distribute binaries; languages typically already offer their own mechanisms for installing executables (`go install`, `npm isntall -g`, `pip install`, etc), and so do OS-level tools (apt-get, rpm, winget, yum, brew). Also: a tarball isn't a sufficient way to distribute executables; some executables need to link to libraries (e.g. glibc), and others require a runtime (e.g. Node.js). It's better to delegate this task to something else.

Some additional thoughts:

* Right now, the message passing API is held outside of pkl-core. We did this to reduce the number of dependencies (no msgpack), and to minimize the amount of core that gets shipped in our core library. This feature implies that much of our message passing logic (e.g. `MessageTransport`) gets moved to pkl-core. However, I think we want this anyway--we've talked about moving BinaryEvaluator into pkl-core, and providing a serializer/deserializer to our Java library users (and perhaps even in-language).
* I don't know if we want shelling out to executables to become the norm for Pkl. This feature makes sense to bring feature parity to the CLI, but library users may not want Pkl to spawn executables. However, it still makes sense that library users should be _able_ to configure Pkl to do this if the CLI can do this too.

HT154:
I've taken another pass on this. Here are the major changes:
* Moved URI-based reader executables to future directions
* Addressed exec-per-read and up-front scheme->reader registration under alternatives considered
* Added significant detail to the detailed design section

spices/SPICE-0009-external-readers.adoc
bioball: note: see comments [here](https://github.com/apple/pkl-evolution/pull/10#pullrequestreview-2227333322) and [here](https://github.com/apple/pkl-evolution/pull/10#issuecomment-2278850140) about why not distribute binaries using a package

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,319 @@
...
+=== Overall Flow
+
+1. User requests an evaluator (via Java API, Message Passing API, or CLI) with an external reader.
+2. Entrypoint (client Java code, `CliEvaluator`, or `Server`) instantiates an `ExternalReader`.
+3. The `ExternalReader` is started, which starts the configured child process and sends a `DiscoverReadersRequest` message.
+4. A `DiscoverReadersResponse` message is received (within some timeout).
+5. The entrypoint requests resource readers (`ExternalResourceReader`) and module key factories (`ExternalModuleKeyFactory`) from the `ExternalReader` and uses them to build an evaluator.
+`ExternalResourceReader` and `ExternalModuleKeyFactory` function similarly to the existing `ClientResourceReader` and `ClientModuleKeyFactory`, implementation may be reusable entirely.
+6. Evaluation begins.
+7. A resource or module read is encountered.
```
bioball, line 84: I think steps 2 - 5 should be inserted here. We should defer spawning a child process to when it's actually needed.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,319 @@
...
+type ExternalReader struct {
+	Executable string
+	Arguments []string
+}
+----
+
+A new `ExternalReaderRuntime` type will be introduced to implement the child process message passing interface.
+It makes sense to expand the existing libraries to add this functionality as much of the message passing infrastructure and types for implementing resource and module readers can be reused.
+An `ExternalReaderRuntime` is configured with zero or more `ResourceReader` instances and zero or more `ModuleReader`.
+When started, the runtime will consume messages from standard input, dispatch calls to the configured readers, and send responses over standard output.
```
bioball, line 202: Probably better to be passed a reader/writer interface

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,319 @@
...
+3. The `ExternalReader` is started, which starts the configured child process and sends a `DiscoverReadersRequest` message.
+4. A `DiscoverReadersResponse` message is received (within some timeout).
+5. The entrypoint requests resource readers (`ExternalResourceReader`) and module key factories (`ExternalModuleKeyFactory`) from the `ExternalReader` and uses them to build an evaluator.
+`ExternalResourceReader` and `ExternalModuleKeyFactory` function similarly to the existing `ClientResourceReader` and `ClientModuleKeyFactory`, implementation may be reusable entirely.
+6. Evaluation begins.
+7. A resource or module read is encountered.
+8. The appropriate `Read*Request` message is sent to the child process, the corresponding `Read*Response` message is awaited.
+9. Evaluation continues.
+10. Evaluation completes.
+11. The `ExternalReader` is closed, terminating the child process.
```
bioball, line 88: Some detail here would be good. Does it send SIGKILL/SIGTERM?

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,319 @@
...
+* `ExternalModuleKeyFactory` - a `ModuleKeyFactory` implementation similar (identical?) to `ClientModuleKeyFactory`.
+* `ExternalModuleKey` - a `ModuleKey` implementation similar (identical?) to `ClientModuleKey`.
+* `ExternalReader` - manages the lifecycle of child processes.
+    ** Explicit `open`/`close` methods to manage child process lifecycle.
+    ** Opening an `ExternalReader` spawns the subprocess, sets up the `MessageTransport`, sends a `DiscoverReadersRequest` message, and awaits a `DiscoverReadersResponse` response.
+    ** Methods to procure `ExternalResourceReader` and `ExternalModuleKeyFactory` instances.
+
+This proposal requires that the message passing API functionality move out of pkl-server.
+A new project pkl-msgapi will be created to contain the new APIs and the core messaging code current part of pkl-server (`pkl-server/src.main/kotlin/org.pkl.server/Message*.kt`).
+The pkl-msgapi project will be depended on by pkl-cli, pkl-server, and Java API clients using external readers.
```
bioball, line 104: Might be better to just move the message passing API into pkl-core altogether.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,319 @@
...
+
+1. User requests an evaluator (via Java API, Message Passing API, or CLI) with an external reader.
+2. Entrypoint (client Java code, `CliEvaluator`, or `Server`) instantiates an `ExternalReader`.
+3. The `ExternalReader` is started, which starts the configured child process and sends a `DiscoverReadersRequest` message.
+4. A `DiscoverReadersResponse` message is received (within some timeout).
+5. The entrypoint requests resource readers (`ExternalResourceReader`) and module key factories (`ExternalModuleKeyFactory`) from the `ExternalReader` and uses them to build an evaluator.
+`ExternalResourceReader` and `ExternalModuleKeyFactory` function similarly to the existing `ClientResourceReader` and `ClientModuleKeyFactory`, implementation may be reusable entirely.
+6. Evaluation begins.
+7. A resource or module read is encountered.
+8. The appropriate `Read*Request` message is sent to the child process, the corresponding `Read*Response` message is awaited.
```
bioball, line 85: In our evaluator API, the order of readers matters; when reading a module, Pkl will try the first reader, then the second, then the third, and so on. In Java, and in jpkl, the last reader is `GenericUrl`, which tries to open all URIs as URLs (for example, it can handle `ftp`).

If an external reader is provided, what is its precedence compared to other readers?

HT154: That's a great question!

Do you think it would be reasonable to do something like this? From highest to lowest precedence:
1. Any pkl-server `Client` readers - these are currently first and it seems important to preserve this for compatibility
2. Any external readers configured via CLI / Message passing API, in order specified
3. Any external readers configured via PklProject, in order specified
4. All of Pkl's built-in readers in the current order

The current behavior of pkl-server client readers actually surprised me a little.
If there's concern around overriding Pkl's built-in readers, it might make sense to instead do something like this:
1. Any pkl-server `Client` readers
2. All of Pkl's built-in readers in the current order, except `GenericUrl`
3. Any external readers configured via CLI / Message passing API, in order specified
4. Any external readers configured via PklProject, in order specified
5. `GenericUrl`

spices/SPICE-0009-external-readers.adoc
HT154: Great idea! Updated

spices/SPICE-0009-external-readers.adoc
HT154: Sounds good, updated! I also realized that allowing configuration of external readers via PklProject would most likely necessitate these being in pkl-core.

spices/SPICE-0009-external-readers.adoc
HT154: Updated!

spices/SPICE-0009-external-readers.adoc
HT154: So that I'm sure I'm understanding correctly, instead of declaring the scheme->command process up-front, instead, the first time the `ExternalModuleKeyFactory.create` is called, launch the process and perform the discovery. That would be really ideal, but the challenge I'm seeing is that `ResourceReader` must implement `getUriScheme` so the schemes must be known at `ResourceManager` init time. Am I understanding you correctly? If so, any recommendations on how to handle this for resource readers?

bioball: Sorry, I'm simply saying that we should defer spawning a process to when that process is actually needed, rather than spawn it prior to the start of evaluation. That way, Pkl only pays the cost of spawning that process for programs that actually read from a resource.

spices/SPICE-0009-external-readers.adoc
bioball: > 1. Any pkl-server `Client` readers - these are currently first and it seems important to preserve this for compatibility
> 2. Any external readers configured via CLI / Message passing API, in order specified
> 3. Any external readers configured via PklProject, in order specified
> 4. All of Pkl's built-in readers in the current order

This makes sense to me! But note that the evaluator doesn't actually about `PklProject` is. `CreateEvaluatorRequest` expects that the client that interacts with to evaluate a PklProject beforehand, and then make its own decisions about what to do with the evaluator settings found there. Our various language bindings have methods that mimick the precedence used by pkl-cli, which is similar to what you've laid out here.

Also: we should consider adding this to `~/.pkl/settings` too.
This enables the use-case of installing a reader globally, which then enables Pkl to use that reader anywhere.

> If there's concern around overriding Pkl's built-in readers, it might make sense to instead do something like this:

We don't recommend doing this, but Pkl intentionally allows this for those that really need to monkey patch Pkl.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,335 @@
...
+== Detailed design
+
+To avoid terminology confusion with the existing language binding message passing model, the `pkl` process will remain known as the "server" and the reader process will be known as the "client", despite all requests in this proposal being initiated by the `pkl` process.
+
+=== Overall Flow
+
+1. User requests an evaluator (via Java API, Message Passing API, or CLI) with an external reader.
+2. Entrypoint (client Java code, `CliEvaluator`, or `Server`) instantiates an `ExternalReader`.
+3. The `ExternalReader` is started, which starts the configured child process and sends a `DiscoverReadersRequest` message.
+4. A `DiscoverReadersResponse` message is received (within some timeout).
```
bioball, line 80: Thinking about this a little more--what's the use-case for wanting discovery?

Asides from our `GenericUrl` catch-all reader in Java, I'm not really aware of other use-cases for having module readers handle unknown schemes ahead of time. The rest of our readers handle a statically known set of schemes.

The API of our message passing API is more limited, and requires that each reader's scheme be known ahead of time. We did this because it wasn't really clear if we ever needed a more "generic" reader model that the Java side has.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,335 @@
...
+$ pkl eval <module> --external-resource-reader ldap=pkl-ldap --allowed-resources prop:,ldap:
+username = "john_appleseed"
+email = "appleseed@example.com"
+----
+
+This approach raises a few questions:
+
+* Should declaring an external reader automatically allow reads for its scheme?
+Declaring explicit allowed resources should probably disable this behavior.
+* What happens when a reader doesn't support the scheme it is declared for?
```
bioball, line 335: What does this mean? Is this when the reader sends back an error message in its `Read {Module,Resource} Response`? If so, that should be bubbled back up as a Pkl Error during evaluation.

spices/SPICE-0009-external-readers.adoc
bioball: Another benefit of requiring users to declare schemes ahead of time is that we don't need to spawn these external processes unless executing an `import` or `read` that uses their scheme. This can be important if, say, we allow external readers to be configured globally per machine (e.g. by adding them to `~/.pkl/settings.pkl`).

HT154: The biggest motivation is user experience.

Here are three takes on discovery:

## Full discovery (as proposed)

The supported schemes (module and resource) and their local/globbable/hierarchical flags can be discovered purely with knowledge on how to invoke the process. This yields the most "on rails" user experience:

```console
--external-reader pkl-ldap
# or
--external-reader pkl-myproto
```
Or in PklProject
```pkl
evaluatorSettings {
  externalReaders {
    new { executable = "pkl-ldap" }
  }
}
// or
evaluatorSettings {
  externalReaders {
    new { executable = "pkl-myproto" }
  }
}
```

## No discovery at all

In a world without any discovery, the scheme mapping and local/globbable/hierarchical flags must be configured up front. The config must look something like this:
```console
--external-resource-reader ldap=pkl-ldap --external-resource ldaps=pkl-ldap
# or
--external-resource-reader myproto=pkl-myproto --external-resource-globabable myproto --external-resource-hierarchical myproto --external-resource-local myproto --external-module-reader myproto=pkl-myproto --external-module-globabable myproto --external-module-hierarchical myproto --external-module-local myproto
```
Or in PklProject
```pkl
evaluatorSettings {
  externalResourceReaders {
    new { scheme = "ldap"; executable = "pkl-ldap" }
    new { scheme = "ldaps"; executable = "pkl-ldap" }
  }
}
// or
evaluatorSettings {
  externalResourceReaders {
    new { scheme = "myproto"; executable = "pkl-myproto"; globbable = true; hierarchical = true; local = true }
  }
  externalModuleReaders {
    new { scheme = "myproto"; executable = "pkl-myproto"; globbable = true; hierarchical = true; local = true }
  }
}
```

This is a ton of config for users to get wrong. Reader implementations already know these properties about themselves and should be able to answer for them.

## Minimal discover

A compromise here is to require manual scheme mapping still, but launch the process on first read and perform discovery of a reader process's local/globbable/hierarchical flags for that module/resource scheme. Config might look like this;
```console
--external-resource-reader ldap=pkl-ldap --external-resource-reader ldaps=pkl-ldap
# or
--external-resource-reader myproto=pkl-myproto --external-module-reader myproto=pkl-myproto
```
Or in PklProject
```pkl
evaluatorSettings {
  externalResourceReaders {
    new { scheme = "ldap"; executable = "pkl-ldap" }
    new { scheme = "ldaps"; executable = "pkl-ldap" }
  }
}
// or
evaluatorSettings {
  externalResourceReaders {
    new { scheme = "myproto"; executable = "pkl-myproto" }
  }
  externalModuleReaders {
    new { scheme = "myproto"; executable = "pkl-myproto" }
  }
}
```

I still don't think this is ideal from a user experience perspective, but it does allow very straightforward deferral of process spawn until first read.

spices/SPICE-0009-external-readers.adoc
HT154: Taking the `pkl-ldap` reader described here as an example: it answers for resources with the `ldap` and `ldaps` schemes. If a user makes a mistake and configures the read eg. as `--external-resource-reader postgresql=pkl-ldap` then the reader process can receive messages with a scheme it doesn't support. The discovery-driven design can return the native Pkl "no reader registered for scheme error", while explicit registration means the request is still sent to the external reader where the returned error is dependent on the client-side implementation which hopefully can handle this gracefully and return a helpful error.

spices/SPICE-0009-external-readers.adoc
HT154: One advantage of the second approach (combined with the `ResourceReaderFactory` approach from https://github.com/apple/pkl/pull/647) is that no external reader process can be spawned unless a read is attempted for a protocol that Pkl does not implement its own reader native for.

> Also: we should consider adding this to ~/.pkl/settings too.
> This enables the use-case of installing a reader globally, which then enables Pkl to use that reader anywhere.

I had explicitly omitted that from this proposal as I couldn't think of a use case for it. Do you have a use case in mind for this?

HT154: > Pkl intentionally allows this for those that really need to monkey patch Pkl

This might be an interesting route toward implementing authentication for HTTP(S) resources in lieu of specific handling for this in the language!

spices/SPICE-0009-external-readers.adoc
HT154: I think I've successfully convinced myself that the minimal discovery mechanism has enough benefits with few enough downsides to be the correct path forward here. I'm working on revising the SPICE to reflect that.

This choice raises some more questions:
* Should explicitly registering a reader for a scheme automatically allow resource/module reads for that scheme?
  * This would be ignored if explicit allow-lists are configured, which matches the behavior for built-in readers
  * This is a nice tradeoff for the slightly more involved configuration in simple cases
* Should any kind of warning be produced when built-in readers are overridden?
  * I can see a lot of potential for nasty surprises when doing this, especially if it becomes possible to configure this via `~/.pkl/settings.pkl`.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,343 @@
...
+`ResourceReaders.External` and `ModuleKeyFactories.External` function almost identically to the existing `ClientResourceReader` and `ClientModuleKeyFactory` and pkl-server will be updated to reuse these implementations where possible.
+4. Evaluation begins.
+5. A resource or module read is encountered.
+6. Upon first read only:
+1.. The `ExternalProcess` is started, which starts the configured child process.
+2.. The child process is sent a `Discover(Module|Resource)ReaderRequest` message.
+3.. A `DiscoverReadersResponse` message is received (within some timeout).
+7. The appropriate `Read*Request` message is sent to the child process, the corresponding `Read*Response` message is awaited.
+8. Evaluation continues.
+9. Evaluation completes.
```
HT154, line 95: Updated with deferred process spawn

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,343 @@
...
+10. The `ExternalProcess` is closed, the child process is sent SIGTERM.
+11. If the child process has not terminated after some timeout (3 seconds?), it is forcefully stopped via SIGKILL.
+
+=== Java API
+
+New APIs:
+
+* `ResourceReaders.External` - a `ResourceReader` implementation identical to `ClientResourceReader`.
+* `ModuleKeyFactories.External` - a `ModuleKeyFactory` implementation similar to `ClientModuleKeyFactory`.
+* `ModuleKeys.External` - a `ModuleKey` implementation identical to `ClientModuleKey`.
```
HT154, line 105: Updated to note exactly which parts of pol-server can be moved, ported, and reused entirely

HT154 (commented):
I've updated the proposal to require explicit scheme->command registration and the "minimal discovery" approach described here: https://github.com/apple/pkl-evolution/pull/10#discussion_r1764308461

I've also gotten far enough long in implementation to update wording in some places to be more certain.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,343 @@
...
+
+In this example, the external reader may provide both `ldap:` and `ldaps:` schemes.
+External readers are registered by their scheme, so to support both schemes both would need to be specified on the command line:
+[source,text]
+----
+$ pkl eval <module> --external-resource-reader ldap=pkl-ldap --external-resource-reader ldaps=pkl-ldap
+----
+
+Registering an external reader for a scheme automatically adds that scheme to the default allowed modules/resources.
+As with Pkl's built-in module and resource schemes, setting explicit allowed module or resources overrides this behavior and appropriate patterns must be specified to allow use of external readers.
```
HT154, line 75: Note on allowing registered external readers by default now that explicit scheme registration is needed.

spices/SPICE-0009-external-readers.adoc
stackoverflow: Agree with Dan here that having a statically defined mapping of external readers is better. Not only is much more performant (Waiting on multiple readers for a scheme listing response may tank performance, specially for small configs), it's also more readable: you can see directly in the cli/setting.pkl/PklProject exactly which schemes are supported, no needed to look into the readers themselves.

I disagree that making it explicit makes the configuration more complicated. The user has to know which schemes the reader supports anyway, because they have to write the read/import string with that scheme.

Another benefit: let's say a reader supports three schemes but I just want to use one in my project. I'd just add that one and I don't have to worry about other schemes sneaking in into my code, making it safer.

bioball: I think the "minimal" approach makes sense. Flags like hierarchical/globbable/local are implementation details of a reader, so it's kind of nice that users don't need to configure it explicitly. However, the scheme is an important aspect of the reader that both implementers and users should be aware of, so it makes sense to me that they need to provide it to the evaluator.

> Should explicitly registering a reader for a scheme automatically allow resource/module reads for that scheme?

I think it depends; for the CLI, `pkl:Project` and `pkl:settings`, it might make to add its scheme to the default set of allowed resources/modules if not otherwise configured:

But I don't think this makes sense for `CreateEvaluatorRequest`, nor for the evaluator builder. Those are lower level, and users can build helpers easily to configure these.

We can also add helpers in pkl-swift and pkl-go to set them. We already have this today with adding module/resource readers.

> Should any kind of warning be produced when built-in readers are overridden?

The flip side is that can be quite annoying for users that know what they're doing and intentionally set them. I don't feel strongly about this right now, though.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,345 @@
...
+  /// A number identifying this request
+  requestId: Int
+
+  /// Client-side resource reader spec.
+  ///
+  /// Null when the external process does not implement the requested scheme.
+  /// [ClientResourceReader] is defined at https://pkl-lang.org/main/current/bindings-specification/message-passing-api.html#create-evaluator-request
+  spec: ClientResourceReader?
+}
+----
```
bioball, line 164: Since the scheme is known, it's not so much of a discovery anymore. How about let's call this `InitializeModuleReader` and `InitializeResourceReader`?

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,345 @@
...
+externalResourceReaders: Mapping<String, ExternalReader>?
+
+class ExternalReader {
+  /// May be specified as an absolute path to an executable
+  /// May also be specified as just an executable name, in which case it will be resolved according to the PATH environment variable
+  executable: String
+
+  /// Command line arguments that will be passed to the reader process
+  arguments: Listing<String>
+}
```
bioball, line 180: I think let's be consistent with `clientModuleReaders` and `clientResourceReaders`, and make each  of these a `List<ExternalReader>`.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,345 @@
...
+----
+
+=== CLI
+
+New `--external-resource` and `--external-module` CLI argument will be added to configure external readers.
+The arguments can be provided multiple times to configure multiple external readers.
+The arguments are passed as `=`-delimited key-value pairs where the key is the reader's URI scheme.
+The argument values may be passed as space-separated strings where the first element becomes `executable` and any remainder becomes `arguments`.
+
+TBD: It might be best if the argument value is link:https://docs.python.org/3/library/shlex.html#shlex.split[shlex'd] instead of split to support passing arguments to the reader process that contain spaces.
```
bioball, line 190: What does this look like?

spices/SPICE-0009-external-readers.adoc
```diff
@@ -0,0 +1,345 @@
...
+5. A resource or module read is encountered.
+6. Upon first read only:
+1.. The `ExternalProcess` is started, which starts the configured child process.
+2.. The child process is sent a `Discover(Module|Resource)ReaderRequest` message.
+3.. A `DiscoverReadersResponse` message is received (within some timeout).
+7. The appropriate `Read*Request` message is sent to the child process, the corresponding `Read*Response` message is awaited.
+8. Evaluation continues.
+9. Evaluation completes.
+10. The `ExternalProcess` is closed, the child process is sent SIGTERM.
+11. If the child process has not terminated after some timeout (3 seconds?), it is forcefully stopped via SIGKILL.
```
bioball, line 97: Steps 10 and 11 should be when the evaluator itself closes, rather than when evaluation completes. This way, we can re-use an external process for multiple evaluations.

spices/SPICE-0009-external-readers.adoc
HT154: Let's say that an external reader is configured like so:
```
--external-resource myScheme='myExecutable arg1 "arg2 with spaces"'
```

A naive white space trim+split yields
```pkl
executable = "myExecutable"
arguments {
  "arg1"
  "\"arg2"
  "with"
  "spaces\""
}
```

While a "shlex" over the input behaves much more similarly to what a user might expect and results in
```pkl

arguments {
  "arg1"
  "arg2 with spaces"
}
```

So without the shlex operator for the CLI args it's actually not possible to express the same configs on the CLI as it is via PklProject.

spices/SPICE-0009-external-readers.adoc
HT154: Updated

spices/SPICE-0009-external-readers.adoc
HT154: The behavior you suggest is how the implementation currently works, updated the language here accordingly.

spices/SPICE-0009-external-readers.adoc
```diff
@@ -118,28 +119,18 @@ Four new message types will be added:
...
 /// Type: Server Request
-class DiscoverModuleReaderRequest {
-  /// A number identifying this request
+class InitializeModuleReaderRequest {
+  /// A number identifying this request.
   requestId: Int
 
-  /// The scheme of the resource to discover the spec for
+  /// The scheme of the resource to initialize.
   scheme: String
```
bioball, line 127: The reader should already know what schemes it should read when it receives a `Read*Request`. Do we need `scheme` here?

HT154: The initialize messages only request the spec for a single scheme at a time. It would be possible to do an "all at once" sort of thing like the original proposal, but this method is actually super convenient from an implementation standpoint as it enables using the battle-tested `ConcurrentHashMap` response memoization already widely employed by the message passing API. My initial pass at the implementation that did the "all at once" discovery needed some messy locking; it would be great to pass the buck to the standard library instead.

PR #11, opened 2024-08-15
SPICE-0010: Overhauled mapping/listing typechecks

bioball:
This addresses the following issues:

https://github.com/apple/pkl/issues/405
https://github.com/apple/pkl/issues/406

The implementation for this SPICE can be found here: https://github.com/apple/pkl/pull/628

### DIFF
--- a/spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
+++ b/spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
@@ -209,11 +209,12 @@ people: Listing<Person> = new {
 <1> A type cast is added to each member.
 <2> The overall result is casted to `Listing`.
 
-=== Surrogate objects
+[#delegating-objects]
+=== Delegating objects
 
 Underneath the hood, casting a listing or mapping returns a _new_ listing/mapping.
 
-This new object is a surrogate of the original object.
+This new object delegates to the original object.
 It behaves just like the original, except its members receive type casts too.
 This type cast may or may not throw an error.
 
@@ -247,7 +248,7 @@ listing1 = (listing0) {
 listing2: Listing<Int> = listing1
 ----
 
-`listing2` is a surrogate of `listing1`, so its parent is still `listing0`.
+`listing2` delegates to `listing1`, so its parent is still `listing0`.
 Thus, `super[0]` is a lookup of the first member of `listing0`.
 
 === Operators: Type cast (`as`) vs. Type test (`is`)
@@ -269,7 +270,7 @@ isPeople = people is Listing<Person> // <1>
 ----
 <1> Each member of `people` is evaluated, and asserted to be a `Person`.
 
-On the other hand, the `as` operator is a type cast, and it behaves just like a typecheck does as described in xref:_surrogate_objects[surrogate objects].
+On the other hand, the `as` operator is a type cast, and it behaves just like a typecheck does as described in xref:delegating-objects[delegate objects].
 
 [source,pkl]
 ----
@@ -280,7 +281,7 @@ local people = new Listing {
 
 myPeople = people as Listing<Person> // <1>
 ----
-<1> `myPeople` is a different object an is a surrogate of `people`.
+<1> `myPeople` is a different object and delegates to `people`
 
 === `new` with explicit parent
 
@@ -356,10 +357,12 @@ value3: List<Listing<String>>|Set<String> // <1>
 ----
 <1> `List` is _shallow_ checked; the `Listing` object itself is evaluated, but its members are not.
 
-=== Optimizations: `unknown`, `Any`, un-parameterized types
+=== Optimizations
+
+==== `unknown`, `Any`, un-parameterized types
 
 Mappings/Listings with type parameter `unknown` or `Any` do not need to have their members casted.
-Therefore, the type cast of a value to `Listing<unknown>` can return the same in-memory object rather than return a xref:_surrogate_objects[surrogate].
+Therefore, the type cast of a value to `Listing<unknown>` can return the same in-memory object rather than return a xref:delegating-objects[delegating object].
 
 [source,pkl]
 ----
@@ -367,7 +370,21 @@ myListing: Listing<unknown> = someListing // <1>
 ----
 <1> `listing` and `someListing` are the same in-memory object
 
-Un-parameterized `Mapping` and `Listing` objects also do not need to return surrogate objects. `Listing` is sugar for `Listing<unknown>`.
+Un-parameterized `Mapping` and `Listing` objects also do not need to return delegating object. `Listing` is sugar for `Listing<unknown>`.
+
+==== Same typechecks
+
+Two types that are the same typecheck does not need return a <<delegating-objects,delegating object>>.
+
+In the following snippet:
+
+[source,pkl]
+----
+l1: Listing<String>
+l2: Listing<String> = l1
+----
+
+`l1` and `l2` have the same type, so we do not need to introduce a new delegating object.
 
 == Compatibility
 
--- a/spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
+++ b/spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
@@ -24,7 +24,7 @@ hidden birds: Listing<Bird> = new {
   new { name = "Warbler" }
 }
 
-numBirds = birds.length
+numBirds = birds[0]
 //         ^^^^^
 ----
 
--- a/spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
+++ b/spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
@@ -1,0 +1,432 @@
+= Overhauled Mapping/Listing Typechecks
+
+* Proposal: link:./SPICE-0010-overhauled-mapping-listing-typechecks.adoc[SPICE-0010]
+* Status: TBD
+* Implemented in: Pkl TBD
+* Category: Language
+
+== Introduction
+
+During typechecks for Listings and Mappings, retain laziness of their object members.
+
+== Motivation
+
+Currently, in order to typecheck a listing or mapping, Pkl will shallow-force its members.
+
+For example, in the following snippet:
+
+[source,pkl]
+----
+class Bird { name: String }
+
+hidden birds: Listing<Bird> = new {
+  new { name = "Osprey" }
+  new { name = "Warbler" }
+}
+
+numBirds = birds.length
+//         ^^^^^
+----
+
+During execution of `birds` (as highlighted by the carets), Pkl will execute both members of `birds`, including `new { name = "Warbler" }`, even though neither of the members are used in the result
+
+In psuedocode, Pkl executes the following instructions:
+
+[source]
+----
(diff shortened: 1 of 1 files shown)


spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
```diff
@@ -0,0 +1,432 @@
...
+listingOfNumbers: Listing<Int> = listing
+----
+
+`"hello"` is not an `Int`, so this program should fail.
+Because typechecking is deferred, this failure should not occur until `listing[0]` is evaluated.
+
+Effectively, `listing` and `listingOfNumbers` are _different_ in-memory objects; one which has type assertions at the end of each member, and one which does not.
+This difference is opaque to users; these two objects have the same hash code, and same equality semantics.
+
+In the amends chain, `listingOfNumbers` has the same parent as `listing`. It delegates its member lookups to `listing`, and adds a typecheck at the end of each lookup.
```
translatenix, line 235: @bioball Isn’t this an undesirable breaking change? If I’m not mistaken, introducing a delegation step means that overridden values are no longer reflected in dependent values. Same for the `as` operator.

bioball: Not totally sure what you mean. Can you provide a snippet that you think is breaking?

BTW: this PR is already merged into main, so feel free to play around with it and see if there is a regression here in Pkl's late-binding semantics.

bioball: There's an optimization that we make here: if the target type results in the same typecheck, it doesn't do any delegation. So, in this case, there's still just one mapping. See https://github.com/apple/pkl/blob/a7cc09892533b12a80c434937544406b18ae24b2/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java#L1621-L1623 for implementation details!

bioball: This is the delegation that's happening:

```
m1: Mapping = new {
  ["name"] = "Bob"
  ["name2"] = this["name"]
}

m2: Mapping<String, String> = (m1) { ["name"] = "Marley"  }

// if m2["name2"] simply delegated to m1["name2"], I'd expect "Bob" here
name2 = m2["name2"] // "Marley"
```

translatenix: I'm still trying to wrap my head around how the delegation works:

```
m1: Mapping = new {
  ["name"] = "Bob"
  ["name2"] = this["name"]
}

// apparently, `m2 = m1 as Mapping<String, String>` isn't equivalent
// to the following. Otherwise, late binding would break (see m3).
m2 = new Mapping {
  ["name"] = m1["name"] as String 
  ["name2"] = m1["name2"] as String
}

m3 = (m2) {
  ["name"] = "Marley"  
}
```

bioball: Don't know if this is any more helpful, but... the lookup on `name` isn't `m1["name"] as String`; it's `((m1) { ["name"] = "Marley" })["name"] as String`.

translatenix: After checking the implementation, I think I understand what's going on:
The new object resulting from `m1 as Mapping<String, String>` doesn't really define new members that delegate to `m1`'s members.
Instead, the new object knows its value type, and `as String` happens as part of member access.

spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
```diff
@@ -0,0 +1,432 @@
...
+
+Amended members of a `Listing` or `Mapping` do not receive type casts.
+
+[source,pkl]
+----
+listing: Listing<Int> = new { 1; 2; 3 }
+
+listing2 = (listing) { "hello" } // <1>
+----
+<1> Okay; result is `new Listing { 1; 2; 3; "hello" }`
```
translatenix, line 310: Conceptually, this makes sense to me:
We append a string to a Listing<Int> and get a listing containing ints and strings.

However, the following makes less sense to me:
```
local listing1: Listing<Int> = new { 1 }
local listing2: Listing<String> = listing1
local listing3 = (listing2) { "2" } // append a string to a Listing<String>
first = listing3[0] // we got an int (and no error)
```

bioball: Hm, there must be a bug, then. I would expect this to error. Good catch!

spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
bioball: Yup, that's right!

spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
bioball: Fix: https://github.com/apple/pkl/pull/725

spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
```diff
@@ -0,0 +1,432 @@
...
+[source]
+----
+assert value instanceof Listing
+for elem in value.members():
+  assert elem.eval() instanceof Bird
+----
+
+There are two underlying issues with this approach.
+
+=== Performance cost
```
translatenix, line 44: Note that the current implementation of surrogate objects is itself quite expensive: It adds 5 fields to, and allocates 2 empty collections for each instance of, class `VmListing` and `VmMapping` (via a common superclass). These classes aren't only used for surrogate objects but also for regular listings/mappings. Additionally, case distinctions for surrogate objects are added throughout the codebase.

One way to improve on the current implementation is to move towards an object model that supports different internal representations for the same type of Pkl object. For example, different representations could be used for small, large, and surrogate listings/mappings.

spices/SPICE-0010-overhauled-mapping-listing-typechecks.adoc
```diff
@@ -0,0 +1,432 @@
...
+isMyFunc = myFunc is (Int) -> Int // <2>
+----
+<1> Fine; we can wrap the provided function with a new function that asserts the types of the parameter and return value
+<2> This is impossible to check, because Pkl types do not have the concept of subtypes, and runtime functions do not retain their type signatures.
+
+To be sound, this operation should throw when evaluated.
+
+== Alternatives considered
+
+N/A
```
translatenix, line 432: Have you considered to preserve the element type of `Listing<X>` when the listing is amended? Although a breaking change, this doesn't break the common case `xs: Listing<X>` where  `xs` is repeatedly amended (because the property type stays `Listing<X>`). And it's much more space/time efficient because it doesn't require adding another surrogate after every amend.

bioball: We've thought about this in the past, but this would be quite limiting on the usage side.

For example:

```pkl
local dogs: Listing<Dog> = new {
  new { breed = "Greyhound" }
}

animals: Listing<Animal> = (dogs) {
  new Horse { ... } // fail; can only add `Dog` here
}
```

translatenix: I wonder how often this occurs in practice, and if something like the following would be acceptable in such cases:

```
animals: Listing<Animal> = new {
  ...dogs
  new Horse { ... }
}
```

"obj-surrogate-obj-surrogate" is quite an expensive (and sad) data structure for the common case.

bioball: FYI: I'm switching terminology to "delegate" for this, because "surrogate" turned out to be kind of a confusing term when discussing this concept internally. Will update docs.

I don't think this is quite so bad for perf. You only get one delegate per typecast, and in the normal case, this happens infrequently.

Idiomatic Pkl code typically looks like this, where there is a schema, and then data that adheres to the schema:

```pkl
// MyPeople
people: Listing<Person>
```

```pkl
// myData.pkl
amends "MySchema.pkl"

people {
  new { name = "Barney" }
}
```

This results in only one delegate when `myData` is eval'd; when `people` is accessed and cast to `Listing<Person>` [^1]. Creating new members within `people` here isn't any more expensive than it is today.

There are two cases that I can think of where this might end up being problematic:

1. There are many collections, rather than many members within a collection. E.g. data that ends up looking like `[[1], [2], [3], [4], [5]]`.
2. A type-cast is executed, where the member type is different, and both the original and the casted object are forced (we end up with two objects, both with their own cached members):
    ```pkl
    people1: Listing<People> = new { ... }
    people2: Listing<People(lastName == "Smith")> = people1
    ```

In any case, I think this is an optimization problem, rather than a problem with the design of the language. Changing this would be a big breaking change, and it would also differ from how this works in most languages (`concat`/`plus` usually accepts contravariant members).

[^1]: Note: `people { ... }` desugars to `people = (super.people) { ... }`, except we do not type-check when accessing `super.people`.

translatenix: Design and implementation should inform each other.

Throwing away all reified type information on every amend is unfortunate from an implementation perspective:
* It is expensive in the common case where a listing is amended and assigned to a property of the same type.
   (Each assignment requires a new delegate.)
* It rules out many potential optimizations, such as using a special representation for `Listing<Int>`.

The current design also means that bugs can go undetected in programs that aren't fully typed.

Fortunately, accepting this SPICE doesn't rule out giving alternative designs a fair shot later on (before 1.0).

bioball: > It is expensive in the common case where a listing is amended and assigned to a property of the same type.
(Each assignment requires a new delegate.)

Can you elaborate here? It would be helpful to see a code snippet that demonstrates this.

> It rules out many potential optimizations, such as using a special representation for Listing<Int>.

I'm not so sure; imagine we have a special representation of `Listing<Int>` called `VmIntListing`:

```java
class ListingAmendNode {
  @Specialization
  VmIntListing eval(VmIntListing parent, long[] members) {
    // we can keep the specialized representation 
  }

  @Fallback
  VmListing eval(IntListing parent, Object[] members) {
    // need to drop back to generic `VmListing` now
  }
}
```

> Fortunately, accepting this SPICE doesn't rule out giving alternative designs a fair shot later on (before 1.0).

Yup! And this is a orthogonal to this SPICE anyway (how typechecking works vs. how amend is typed).

translatenix: > Can you elaborate here? It would be helpful to see a code snippet that demonstrates this.

```
a: Listing<Int> = new { 0 }
b: Listing<Int> = (a) { 1 }
c: Listing<Int> = (b) { 2 }
```
In the common case that an amended listing is assigned back to a property of the same type (or even the same property), one delegate is created for every amend, doubling the number of `VmListing` objects.

> I'm not so sure; imagine we have a special representation of Listing<Int> called VmIntListing:

Right, that's exactly what I imagined. My point is that an optimized `VmIntListing` representation will very likely want the entire prototype chain to consist of `VmIntListing`.

PR #12, opened 2024-10-11
SPICE-0011: Const object members

bioball:
Implementation: https://github.com/apple/pkl/pull/678

### DIFF
--- a/spices/SPICE-0011-const-object-members.adoc
+++ b/spices/SPICE-0011-const-object-members.adoc
@@ -1,0 +1,111 @@
+= Const Object Members
+
+* Proposal: link:./SPICE-0011-const-object-members.adoc[SPICE-0011]
+* Author: https://github.com/bioball[Dan Chao]
+* Status: Review
+* Implemented in: TBD
+* Category: Language
+
+== Introduction
+
+This SPICE proposes allowing object members to be `const`.
+
+== Motivation
+
+Today, the `const` modifier can only be applied to class properties.
+
+However, there are times where this rule is insufficient.
+
+When a module amends another module, it has special rules; top-level properties are considered _object properties_, and not _class properties_.
+
+This means that these properties cannot be declared `const`.
+
+[source,pkl]
+----
+amends "foo.pkl"
+
+const bar = 5 // Pkl Error: modifier `const` cannot be applied to object members
+----
+
+Amending modules can also contain classes and typealiases.
+These members must be declared with the `local` modifier.
+
+However, these members cannot reference their enclosing module's properties, because classes and typealiases can only reference properties that are `const`.
+
+One possible scenario is to define a test fixture.
+
+[source,pkl]
+----
+amends "pkl:test"
+
+import "birds.pkl"
+
+local warbler = "Warbler"
+
+local class Warbler extends birds.Bird {
+  name = warbler // Pkl Error: cannot reference `warbler` from here because it is not `const`
+}
+----
+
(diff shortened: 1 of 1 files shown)


stackoverflow (approved):
Looks good.

spices/SPICE-0011-const-object-members.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+When a module amends another module, it has special rules; top-level properties are considered _object properties_, and not _class properties_.
+
+This means that these properties cannot be declared `const`.
+
+[source,pkl]
+----
+amends "foo.pkl"
+
+const bar = 5 // Pkl Error: modifier `const` cannot be applied to object members
```
holzensp, line 27: Here, the `const`-ness is not the issue. Either `bar` is `const` in `foo.pkl`, in which case it couldn't be overridden, or `bar` doesn't occur in `foo.pkl`, so it can't be added here.

The problem description only makes sense in the context of `local`s, in my mind.

spices/SPICE-0011-const-object-members.adoc
```diff
@@ -0,0 +1,111 @@
...
+local const warbler = "Warbler" // Pkl Error: modifier `const` cannot be applied to object members
+
+local class Warbler extends birds.Bird {
+  name = warbler
+}
+----
+
+== Proposed Solution
+
+To fix this papercut, a object members are allowed to be `const`, but must co-appear with the `local` modifier.
```
holzensp, line 67:
```suggestion
To fix this papercut, object members are allowed to be `const`, but must also be `local`.
```

This phrasing feels very roundabout. One (somewhat pedantic) point is that _members_ not co-appearing with any of their modifiers, but rather modifier `const` requires modifier `local`. That said; `local` is what makes the property an object property to begin with.

If my understanding is correct, this phrasing is there to motivate the restriction of `const` to `local` object members only, i.e. to exclude `new Dynamic { const foo = 5 }`. It seems more direct to rephrase the proposal to be `const local members`, rather than `const object members`. Either way, it should be useful to this SPICE to include a description of the `Dynamic` case (which is the only other object property case I can think of).

spices/SPICE-0011-const-object-members.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+[source,pkl]
+----
+amends "foo.pkl"
+
+const bar = "bar" // Pkl Error: modifier `const` can only be applied to object members that are also `local`.
```
holzensp, line 73: As an illustration of the above; `bar` is only an object member if it is `local`.

spices/SPICE-0011-const-object-members.adoc
bioball: I think I'll just get rid of this block. You're right; it's not describing what the problem actually is.

spices/SPICE-0011-const-object-members.adoc
bioball: We actually treat any property defined on an amending module as an object property, rather than a class property.

See the two branches when we parse a module:

Branch 1:
https://github.com/apple/pkl/blob/4b4d81ba9380a1a47c5cab94e654858dafb64482/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java#L188-L213

Branch 2:
https://github.com/apple/pkl/blob/4b4d81ba9380a1a47c5cab94e654858dafb64482/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java#L215-L234

One parses as class properties, one parses as object properties.

PR #13, opened 2025-01-30
SPICE-0012: URL standard library

bioball:


StefMa:
It seems that "search params" are the correct technical wording. However I think if we would go with just Params it would be better understandable... 🤔 Maybe at least introducing an typealias for this? 🤔

Just some random thoughts 🤷‍♂️ I've never seen `SearchParams` in Java/Kotlin yet (as an Android developer). Maybe others have more use for that wording...

spices/SPICE-0012-url-standard-library-module.adoc
```diff
@@ -0,0 +1,336 @@
...
+/// are percent-encoded.
+///
+/// Follows the rules for the `encodeURI` function as described by
+/// [ECMA-262](https://262.ecma-international.org/5.1/#sec-15.1.3.3).
+///
+/// Facts:
+/// ```
+/// encode("https://example.com/some path/") == "https://example.com/some%20path"
+/// ```
+const external function encode(value: String): String
```
stackoverflow, line 192:
```suggestion
const external function encodeUri(value: String): String
```
`encode` is a bit too generic IMO, and it was hard to understand the difference between it and `percentEncode` at first glance.
Could also be called `encodeForUri`, `encodeUriString`, etc.

spices/SPICE-0012-url-standard-library-module.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+=== Sample usage
+
+URLs can be constructed either by using the parser, or directly by setting fields on the struct.
+
+[source,pkl]
+----
+import "pkl:Url"
+
+myUrl: Url = new { // <1>
```
stackoverflow, line 267: `parse` will throw (or return `null`) on an invalid URL. What happens when you create an invalid URL with `new`? Will it throw?
We could have a `UrlBuilder` that allows you to create a URL with new and call `toUrl(orNull)` to validate it.

stackoverflow (commented):
I Like the design. Left some questions in the comments.

spices/SPICE-0012-url-standard-library-module.adoc
```diff
@@ -0,0 +1,336 @@
...
+----
+module pkl.Url
+
+/// The scheme component.
+scheme: AsciiString
+
+/// The username component.
+///
+/// If the URL does not require a username, set to the empty string.
+username: AsciiString
```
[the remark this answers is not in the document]
bioball: Good question. According to the spec, this field (and several other ones) do not admit null:

https://url.spec.whatwg.org/#url-representation

There are some fields that can be null, but this is not one of them.

Practically, I don't know if this makes any difference.

spices/SPICE-0012-url-standard-library-module.adoc
```diff
@@ -0,0 +1,336 @@
...
+///
+/// Follows the specification in <https://url.spec.whatwg.org/#concept-basic-url-parser>.
+class Parser {
+  /// The base URL, if any.
+  base: Url?
+
+  /// Parses [source] into a URL.
+  ///
+  /// Throws if [source] is an invalid URL.
+  external function parse(source: String): Url
```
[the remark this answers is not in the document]
bioball: Our other parsers don't have a `parseOrNull`, so, this follows the same precedent.

If you need to recover from errors, you can use `test.catch`

spices/SPICE-0012-url-standard-library-module.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+// etc
+
+/// The parsed query as search params.
+hidden fixed searchParams: SearchParams? =
+  if (query != null) SearchParams(query)
+  else null
+
+/// Creates a [SearchParams] from the given form encoded string.
+const function SearchParams(input: String): SearchParams = // etc
```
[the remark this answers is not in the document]
bioball: Good question; I don't think so.

Here is the parsing algorithm: https://url.spec.whatwg.org/#urlencoded-parsing

None of these steps involve failure.

spices/SPICE-0012-url-standard-library-module.adoc
bioball: The module's name is already `Url`, so, usage looks like `Url.encode`. I think `URL.encodeUri` would be a little too wordy.

By the way, one of the suggestions of this SPICE is to prefer "URL" over "URI".

bioball:
@StefMa: this API is taken from the URL spec, see:

https://url.spec.whatwg.org/#urlsearchparams

By the way, Java's stdlib doesn't have a library for parsing the query. However, some other languages do, and call them different things:

* Ruby: [CGI](https://ruby-doc.org/stdlib-2.5.3/libdoc/cgi/rdoc/CGI.html#method-c-parse)
* Dart: [queryParameters](https://api.dart.dev/dart-core/Uri/queryParameters.html)
* JavaScript: [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
* go: [ParseQuery](https://pkg.go.dev/net/url?utm_source=godoc#ParseQuery)

Technically, the query string portion of a URL is just a string (no structure). Representing structured data using `&` and `=` characters is [URL form encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST#url-encoded_form_submission).

KushalP:
Documenting something here that we spoke about: will merging this into `main` require codegen implementations to exist as well?

HT154:
From an ergonomic standpoint, I would hope that Pkl's Url type produces language/stdlib-native URL types in each target language. Pretty sure would require changes in both the code generators and the Unmarshal/Decoder implementations.

bioball:
Good point; I need to fill in some details about language bindings. And, I agree, it should probably turn into the target language's URL type.

HT154:
I wonder if it makes sense to expand `read` to accept `Url` values as well. Many of the places I'm using the experimental URI library today pass the "rendered" URI directly to `read`. This obviously isn't a huge impact, but might be a nice little bit of sugar.

spices/SPICE-0012-url-standard-library-module.adoc
bioball: We can ensure that you cannot construct invalid Urls through property types and constraints.

spices/SPICE-0012-url-standard-library-module.adoc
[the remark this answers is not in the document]
bioball: After chatting a little bit more on this, I'm convinced that we should just have a `parseOrNull` here, updated.

bioball:
> I wonder if it makes sense to expand read to accept Url values as well. Many of the places I'm using the experimental URI library today pass the "rendered" URI directly to read. This obviously isn't a huge impact, but might be a nice little bit of sugar.

We possibly want to make `read()` also only accept string literals in the future. Without this, we can't statically analyze a module and determine all of its dependencies. So, with that in mind, I would rather keep `read()` in its current form.

spices/SPICE-0012-url-standard-library-module.adoc
```diff
@@ -0,0 +1,390 @@
...
+> Standardize on the term URL. URI and IRI are just confusing. In practice a single algorithm is used for both so keeping them distinct is not helping anyone. URL also easily wins the https://trends.google.com/trends/explore?q=url,uri[search result popularity contest].
+
+=== module-level properties
+
+The following make up the properties of the Url module:
+
+.pkl.Url
+[source,pkl]
+----
+module pkl.Url
```
HT154, line 52: How would a URI like `env:USER` be represented by this module?

In go, this is parsed as
```
url.URL{
  Scheme: "env",
  Opaque: "USER",
}
```

bioball: This would be parsed as `new URL { scheme = "env"; path = "USER" }`.

See https://url.spec.whatwg.org/#example-url-components

Actually, I wonder how we can better support opaque URLs. The env URL `env://foo/bar` represents an env var with value `//foo/bar`, and doesn't mean "the host is foo".

HT154: I'd like to proffer Go's URL implementation as a carrot and Swift's as the stick:

### Go
Go's `url.Parse` function does (roughly) this:
1. Split off the fragment, if present, and decode/store it
2. Split off the scheme, if present
3. Split off the query, if present, and store it
4. If a scheme is present and the character immediately following the `:` is _not_ `/`, store the remainder in `Opaque` and return
5. Otherwise, proceed to parse the authority and path

`url.URL.String` produces a string from a URL struct. When `Opaque` is non-empty, the result has form `scheme:opaque?query#fragment`, otherwise it's `scheme://userinfo@host/path?query#fragment`.

In practice, I've found this API to be extremely usable and flexible enough for working with the varied URIs found in the Pkl ecosystem. This may be a good design to reference that departs somewhat from the literal spec in the name of usability.

### Swift

Having worked with pkl-swift, I've found Foundation URL type's lack of similar support makes working with opaque URIs like `env:USER` _significantly_ less ergonomic. Parsing just the `"USER"` out requires error-prone mangling of the full URL string, eg. https://github.com/apple/pkl-swift/blob/c280a0e6324097c429bfc19a2cfe5761ab12bbe9/docs/modules/ROOT/pages/external-readers.adoc?plain=1#L33) to get the equivalent of Go's `url.URL.Opaque`.

### Pkl

Given that opaque URIs are already commonplace in the Pkl ecosystem (and used for several of the runtime's built-in resources), I think having first-class support for them in this proposal is necessary. A similar model to Go's could be adopted, but with added type constraints so ensure that `opaque` is never non-null at the same time as `hostname/`port`/`path`.

A case like `env://foo/bar` (or `env:/foo/bar`) is still tricky (and go parses them "wrong"). I wonder if it would be reasonable to do something like this:
```pkl
opaque: String = "<rendered authority/path>"
```
`Url.toString()` would then use this field and ignore the authority/path field. This would allow direct construction of the `Url` to directly set `opaque` while still maintaining the same behavior as non-opaque URLs. Similarly, during parsing, if an authority/path are found (same criteria as Go) then those fields are populated, otherwise `opaque` is set directly.

Thoughts?

bioball: You actually need to do wrangling too with Go's `url` API for opaque URLs. For example, to turn a URL like `env://foo%20/bar` into `//foo /bar`, you'd need:

```go
func getSchemeSpecificPart(u *url.URL) (string, error) {
  return url.PathUnescape(strings.Split(u.String(), ":")[1])
}
```

Whether a URL is opaque or not is really up to the scheme, and a simple rule like "if the char after the colon is `/`, it's opaque" doesn't feel good enough.

PR #14, opened 2025-03-06
SPICE-0013: Bytes standard library class

bioball:
This adds a new SPICE for working with binary data.

Reference implementation: https://github.com/apple/pkl/pull/1019

### DIFF
--- a/spices/SPICE-0013-bytes-standard-library.adoc
+++ b/spices/SPICE-0013-bytes-standard-library.adoc
@@ -172,7 +172,7 @@ res2 = bytes.getOrNull(2) // <2>
 
 === Concatenation
 
-Two `Bytes` values can be concatenated using `+`, producing a the combined result of both values.
+Two `Bytes` values can be concatenated using `+`, producing the combined result of both values.
 
 [source,pkl]
 ----
--- a/spices/SPICE-0013-bytes-standard-library.adoc
+++ b/spices/SPICE-0013-bytes-standard-library.adoc
@@ -94,9 +94,15 @@ external const function Bytes(from: List<UInt8>|String(isBase64)): Bytes
 
 The string representation of `Bytes` will be `Bytes("<base64-contents>")`.
 
-Bytes can be compared using the `==` and `!=` operators, and concatenated using the `+` operator.
-
+Bytes can be compared using the `==` and `!=` operators.
 All other operators are not compatible with bytes.
+
+==== `Bytes` is a wrapper
+
+The `Bytes` class is a wrapper type, and cannot be used as a byte array.
+`Bytes` cannot be concatenated, nor have individual bytes accessed via subscript, nor iterated over with a for generator.
+
+To work with the underlying bytes, access its `value`.
 
 ==== `String` class
 
--- a/spices/SPICE-0013-bytes-standard-library.adoc
+++ b/spices/SPICE-0013-bytes-standard-library.adoc
@@ -1,0 +1,376 @@
+= Bytes standard library class
+
+* Proposal: link:./SPICE-0013-bytes-standard-library.adoc[SPICE-0013]
+* Author: https://github.com/bioball[Dan Chao]
+* Status: TBD
+* Implemented in: TBD
+* Category: Standard Library
+
+== Introduction
+
+This SPICE introduces a new value type called `Bytes`, which represents binary data.
+
+== Motivation
+
+Pkl currently does not have a binary data type.
+
+This introduces the following limitations:
+
+* `pkl eval` can only produce UTF-8 encoded strings.
+* Asides from values coming from `read()`, it is not possible to compute checksums of binary content.
+* It is harder to interop with configuration formats that allow binary data.
+
+== Proposed Solution
+
+A new class will be added to `pkl:base`, called `Bytes`.
+
+== Detailed design
+
+=== Standard library changes
+
+The following changes will be made to the standard library
+
+==== `Bytes` class
+
+A new class will be added to `pkl:base`, called `Bytes`.
+This will be a new value type, (similar to `Duration` and `DataSize`).
+
+[source,pkl]
+----
+@Since { version = "0.29.0" }
+external class Bytes extends Any {
+  /// The value of these bytes.
+  external value: List<UInt8>
+
+  /// The value represented by these bytes, in base64 encoding.
+  external base64: String(isBase64)
+
+  /// The value represented by these bytes, in hexadecimal encoding.
+  external hex: String
+
+  /// The [MD5](https://en.wikipedia.org/wiki/MD5) hash of these bytes as hexadecimal string.
+  ///
+  /// MD5 is cryptographically broken and should not be used for secure applications.
+  external md5: String
+
+  /// The [SHA-1](https://en.wikipedia.org/wiki/SHA-1) hash of these bytes as hexadecimal string.
+  ///
+  /// SHA-1 is cryptographically broken and should not be used for secure applications.
+  external sha1: String
+
+  /// The [SHA-256](https://en.wikipedia.org/wiki/SHA-2) cryptographic hash of these bytes as
+  /// hexadecimal string.
+  external sha256: String
+
+  /// The first 64 bits of the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) cryptographic hash of
+  /// these bytes.
+  external sha256Int: Int
+
+  /// Converts these bytes into a string using UTF-8 encoding.
+  external function encodeToString(): String
+
+  /// Converts these bytes into a string, encoded using [charset].
+  external function encodeToStringWithCharset(charset: Charset): String
+}
+----
+
+Bytes are built using the `Bytes` constructor method.
+
+[source,pkl]
+----
+/// Creates [Bytes] from either a list of ints, or a Base64 encoded string.
+///
+/// Examples:
+/// ```
+/// // Create Bytes from integer literals
+/// bytes1 = Bytes(List(1, 2, 3, 4))
+///
+/// // Create Bytes from a Base64 encoded string
+/// bytes2 = Bytes("AQIDBA==")
+/// ```
+@Since { version = "0.29.0" }
+external const function Bytes(from: List<UInt8>|String(isBase64)): Bytes
+----
+
+The string representation of `Bytes` will be `Bytes("<base64-contents>")`.
+
+Bytes can be compared using the `==` and `!=` operators, and concatenated using the `+` operator.
+
+All other operators are not compatible with bytes.
+
+==== `String` class
+
+The `String` class will receive the following new members:
+
+[source,pkl]
+----
+class String {
+  /* etc */
+
+  /// Tells if this is a valid base64-encoded string.
+  ///
+  /// Facts:
+  /// ```
+  /// "AQIDBA==".isBase64
+  /// !"hello there".isBase64
+  /// ```
+  @Since { version = "0.29.0" }
+  external isBase64: Boolean
+
+  /// Returns the bytes of this string, encoded using UTF-8.
+  @Since { version = "0.29.0" }
+  function toBytes(): Bytes = toBytesWithCharset("UTF-8")
+
+  /// Returns the bytes of this string, encoded using [charset].
+  @Since { version = "0.29.0" }
+  external function toBytesWithCharset(charset: Charset): Bytes
+}
+----
+
+==== `Charset` typealias
+
+A new typealias called `Charset` will be added to the base module.
+
+[source,pkl]
+----
+@Since { version = "0.29.0" }
+typealias Charset = "UTF-8"|"UTF-16"|"ISO-8859-1"
+----
+
+==== `Resource` class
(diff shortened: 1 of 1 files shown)


spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,376 @@
...
+
+  /// The [MD5](https://en.wikipedia.org/wiki/MD5) hash of these bytes as hexadecimal string.
+  ///
+  /// MD5 is cryptographically broken and should not be used for secure applications.
+  external md5: String
+
+  /// The [SHA-1](https://en.wikipedia.org/wiki/SHA-1) hash of these bytes as hexadecimal string.
+  ///
+  /// SHA-1 is cryptographically broken and should not be used for secure applications.
+  external sha1: String
```
netvl, line 59: I think that the very wide consensus in IT sphere currently is that MD5 and SHA-1 hashes should not be used anymore, so adding them to the new API seems a bit fishy. What is the intended value of having them, especially MD5?

Maybe instead of providing hashes directly on the class level (which is, admittedly, convenient but which is also pretty inflexible) it would make sense to provide a carefully designed module specifically for hashing, which would provide hashing functions instead? It feels like it would be more flexible and consistent, especially if that module will grow with new hash algorithms in future.

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,376 @@
...
+A new class will be added to `pkl:base`, called `Bytes`.
+This will be a new value type, similar to `Duration` and `DataSize`.
+
+[source,pkl]
+----
+@Since { version = "0.29.0" }
+external class Bytes extends Any {
+  /// The value of these bytes.
+  external value: List<UInt8>
+
```
HT154, line 44: I see reference to `toList()` below. What's the difference between `toList()` and `value`?

Also, would it make sense to add a `length` property?

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,376 @@
...
+// From a base64-encoded string
+bytes2 = Bytes("AQIDBA==")
+----
+
+==== Concatenating two `Bytes`
+
+[source,pkl]
+----
+bytes3: Bytes = bytes1 + bytes2
+----
```
HT154, line 286: What about subscript operations? I'd generally expect `bytes1[0]` to return the first byte in the buffer

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,376 @@
...
+==== Modifying bytes
+
+NOTE: Although this is possible, Pkl is not efficient at manipulating byte arrays. If heavy computing is needed here, Pkl is likely not the correct tool for this, and should be implemented in terms of `read()` with a custom module reader.
+
+[source,pkl]
+----
+bytes1: Bytes
+
+bytes2 = Bytes(bytes1.toList().take(4))
+----
```
HT154, line 297: At least one usage I have in mind relies pretty heavily on mutation. Whether it makes sense to build API for this into `Bytes` really depends on the performance of round-tripping through `List`. If that's near-free I think this is fine as-is, but if there is meaningful overhead it'd be great to have some basic mutation APIs.

If there were APIs built for this, I'd like to see something like these:
```pkl
function set(index: UInt, value: UInt8): Bytes
function setRange(start: UInt, end: UInt, value: UInt8): Bytes
function replaceRange(start: UInt, end: UInt, data: Bytes(length == end - start)|List<UInt8>(length == end - start)): Bytes
function map(transform: (Element) -> Result): Bytes
function mapIndexed(transform: (Int, Element) -> Result): Bytes
function mapRange(start: UInt, end: UInt, transform: (Element) -> Result): Bytes
function mapRangeIndexed(start: UInt, end: UInt, transform: (Int, Element) -> Result): Bytes
```

spices/SPICE-0013-bytes-standard-library.adoc
bioball: The sentiment of this design is to provide a way to view into a byte array, not really to manipulate them. That's also why this class is called `Bytes`, rather than `ByteList` or something like that.

If you need to manipulate them, you can convert to list, before going back to `Bytes`.

We can explore making `List<UInt8>` more efficient for this.

spices/SPICE-0013-bytes-standard-library.adoc
bioball: This is sentiment is "I want to use `Bytes` as a list", which, I think, let's have that convo in the thread here: https://github.com/apple/pkl-evolution/pull/14#discussion_r1984168905

BTW: the signature of the map methods are off. Should be either:

```pkl
function map(transform: (Element) -> Result): List<Result>
function mapIndexed(transform: (Int, Element) -> Result): List<Result>
function mapRange(start: UInt, end: UInt, transform: (Element) -> Result): List<UInt8|Result>
function mapRangeIndexed(start: UInt, end: UInt, transform: (Int, Element) -> Result): List<UInt8|Result>
```

Or:

```pkl
function map(transform: (UInt8) -> UInt8): Bytes
function mapIndexed(transform: (Int, UInt8) -> UInt8): Bytes
function mapRange(start: UInt, end: UInt, transform: (Element) -> UInt8): Bytes
function mapRangeIndexed(start: UInt, end: UInt, transform: (Int, Element) -> UInt8): Bytes
```

Which one did you mean?

spices/SPICE-0013-bytes-standard-library.adoc
bioball: > Maybe instead of providing hashes directly on the class level (which is, admittedly, convenient but which is also pretty inflexible) it would make sense to provide a carefully designed module specifically for hashing, which would provide hashing functions instead? It feels like it would be more flexible and consistent, especially if that module will grow with new hash algorithms in future.

We can maybe add a `pkl:crypto` module or something similar in the future, if the need arises. But, these methods are also included because we're migrating the existing `Resource.{sha1,sha256,etc}` methods into `Bytes`.

spices/SPICE-0013-bytes-standard-library.adoc
HT154: Ah, I meant the latter. Got distracted after a copy pasta and forgot to edit those.

spices/SPICE-0013-bytes-standard-library.adoc
bioball: Ah, I went back and forth on these two names, and forgot to update this.

I think `toList()` is probably better here; more consistent with the rest of our API.

bioball: I actually went back on this; I'm keeping this as `value`. This better captures the idea that `Bytes` is just a wrapper, and has an underlying value.

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,420 @@
...
+[source,pkl]
+----
+bytes3: Bytes = bytes1 + bytes2
+----
+
+==== Manipulating bytes
+
+`Bytes` is a wrapper class around an underlying list of bytes.
+
+The underlying bytes can be manipulated by accessing its `value`.
```
stackoverflow, line 292: Why have a `value` property instead of a `toList()` function?
I imagine `value` will create a new VmList either way.
It's also an easier mnemonic to remember when you forgot how to convert to list.

spices/SPICE-0013-bytes-standard-library.adoc
KushalP: As called out here, this functionality exists and is present in the implementation PR: https://github.com/apple/pkl/pull/1019#discussion_r1991869339

> If you need to manipulate them, you can convert to list, before going back to Bytes.

If the desire is to guide users towards this behaviour, should we disable this functionality in `Bytes` or leave it as it is? I'm less concerned with whether subscript operations should/shouldn't exist and what the supported interface looks like for `Bytes` going forwards.

bioball: Right; I will get rid of that behavior in accordance with this SPICE, as long as this behavior is agreed upon.

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,426 @@
...
+
+[source,pkl]
+----
+class FileOutput {
+  /* etc */
+
+  /// The textual rendered output.
+  text: String = renderer.renderDocument(value)
+
+  /// The underlying byte array of [text], in UTF-8 encoding.
```
stackoverflow, line 176:
```suggestion
  /// Defaults to the underlying byte array of [text], in UTF-8 encoding.
```
You can overwrite it, so the doc is misleading.

spices/SPICE-0013-bytes-standard-library.adoc
bioball: I've now renamed this to `toList()`

bioball:
I've updated this document quite a bit.

Initially, I went down the direction of simply making `Bytes` a wrapper class, where access and iteration came from converting it into a list.

However, this was somewhat problematic; rationale detailed in https://github.com/apple/pkl-evolution/pull/14/files#diff-3a286f2e59f0c6b7edd5d791874536f6d1c289dd2f0a9aaf82b9c687db6e83bdR594

I've now updated, so that `Bytes` is itself a sequence of bytes.

spices/SPICE-0013-bytes-standard-library.adoc
bioball: Update here: I've updated this SPICE to make `Bytes` itself a sequences of `UInt8`, and added subscript (see [comment](https://github.com/apple/pkl-evolution/pull/14#issuecomment-2870818716))

The actual changes made can be seen in this commit: https://github.com/apple/pkl-evolution/pull/14/commits/2c712eae4eb8d7220d61179f0a8bf85fce6cdb87

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,643 @@
...
+bytes = Bytes(1, 2, 3, 4)
+res1 = bytes.getOrNull(5) // <1>
+res2 = bytes.getOrNull(2) // <2>
+----
+<1> Result: `null`
+<2> Result: `3`
+
+=== Concatenation
+
+Two `Bytes` values can be concatenated using `+`, producing a the combined result of both values.
```
stackoverflow, line 175:
```suggestion
Two `Bytes` values can be concatenated using `+`, producing the combined result of both values.
```

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,643 @@
...
+
+[source,pkl]
+----
+Bytes(1, 2, 3, 4) + Bytes(5, 6, 7, 8) // <1>
+----
+<1> Result: `Bytes(1, 2, 3, 4, 5, 6, 7, 8)`
+
+=== Generators
+
+`Bytes` can be iterated over by a for-generator, and spread syntax.
```
stackoverflow, line 185: I wonder how useful is that. One could always call `toList()` before spreading or foreaching. It's definitely faster this way, though.

spices/SPICE-0013-bytes-standard-library.adoc
```diff
@@ -0,0 +1,643 @@
...
+
+==== Creating `Bytes`
+
+[source,pkl]
+----
+// From a list
+bytes1 = Bytes(1, 2, 3, 4)
+
+// From a base64-encoded string
+bytes2 = "AQIDBA==".base64DecodedBytes
```
stackoverflow, line 432: Where is `base64DecodedBytes` defined? It doesn't exist in the current `String` class and you didn't specify it as a new property.

stackoverflow (approved):
Look goods. Left some small comments.

spices/SPICE-0013-bytes-standard-library.adoc
bioball: Whoops, added: https://github.com/apple/pkl-evolution/blob/509c11a8b678f3d6f7bd637cbd4456dba107e5d3/spices/SPICE-0013-bytes-standard-library.adoc#string-class

spices/SPICE-0013-bytes-standard-library.adoc
bioball: Update: this got changed back again; `toList()` is back. See https://github.com/apple/pkl-evolution/blob/509c11a8b678f3d6f7bd637cbd4456dba107e5d3/spices/SPICE-0013-bytes-standard-library.adoc#bytes-as-a-wrapper for rationale

spices/SPICE-0013-bytes-standard-library.adoc
bioball: Adding a note here: we haven't included these methods in 0.29,. I'm not opposed to it, but we'd like to see real-world perf issues first.

For now, you'd go t o `List` if transformations are needed.

PR #15, opened 2025-03-07
add spice for the formatter

stackoverflow:


spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+
+The formatter will follow a set of rules. These rules are set in stone and cannot be configured.
+It's important to agree on these rules now, because changing them later would cause a big incompatibility problem
+for users, where their whole codebase would be formatted different due to a rule change.
+
+Some of these rules were taken from our https://pkl-lang.org/main/current/style-guide/index.html[style guide] verbatim,
+others had to be made more strict.
+
+Rules are presented here in order of more generic to less generic.
+Less generic rules (bigger numbers) overrule previously defined ones, if they conflict.
```
madrob, line 92: "more generic to less generic" is hard to mentally parse. "general to specific" is probably more clear.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  bar {
+    baz = "hi"
+  }
+}
+----
+
+==== [0008] Modifiers
+
+Modifiers must be written in the same line as what they are modifying, except if they go over the maximum line limit. +
+Modifiers must be ordered alphabetically.
```
madrob, line 344: could this result in unintuitive reorderings? I think you need to generate a list of all combinations of modifiers and examine that lexicographic ordering groups them appropriately. Otherwise you might end up with X Y property and Y Z property, with mutually exclusive X and Z as things that rally should be in the same place.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+[source, pkl]
+----
+const hidden foo = 1
+
+local open class Foo {}
+----
+
+==== [0009] Class header
+
+The class header must be written in a single line if it doesn't go over the maximum line limit.
```
madrob, line 367: What are the line break rules for `class ReaaaallyLong extends Bar, Foo, Baz, AnotherReallyLong, Etc {}`

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  1000006
+}
+----
+
+==== [0013] Expressions
+
+Expression must either be defined in a single line on in separated lines. +
+If an expression spans multiple lines there must be a new line after the `=`. +
+If the expression is a `new` or amending expression it should start in the same line as the `=`. +
+Operators separated by a new line must be written in the same as the left operand.
```
madrob, line 508: same what?

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+
+.input
+[source, pkl]
+----
+typealias NonEmptyString = String(!isEmpty)
+----
+
+==== [0020] Chained methods and properties
+
+Chains of methods and properties must be separated into a new indented line containing the dot. +
```
madrob, line 711: even if they fit on the same line?

madrob (commented):
I think there should also be formatter:off tags or directives, because sometimes the human will really know better. Maybe it's a structured diagram in the comments, or vertically aligning some tabular data to make it easier to read, or whatever else. But the formatter can't be right 100% of the time.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  -l, --list         Check if the inputs are properly formatted,
+                     printing the file name to stdout in case they are not.
+                     Returns non-zero in case of failure.
+  -r, --replace      Format all the files in place.
+  -f, --format DIR   Format all the files storing the results in the given directory.
+----
+
+=== Java API
+
+The Java API will provide a way to use the formatter programmatically from any JVM language.
```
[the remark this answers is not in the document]
stackoverflow: No. The message passing API is for the core language. This is not part of the core language (it won't land in `pkl-core`), it's tooling.

stackoverflow:
> I think there should also be formatter:off tags or directives, because sometimes the human will really know better. Maybe it's a structured diagram in the comments, or vertically aligning some tabular data to make it easier to read, or whatever else. But the formatter can't be right 100% of the time.

Thanks for the review!

> Maybe it's a structured diagram in the comments

The formatter won't format comments for now (that would require a markdown parser) so this should be fine.

> I think there should also be formatter:off tags or directives...

The idea here is not to make a formatter that prettifies the code, it should just make it canonical. So I don't think supporting a escape hatch is a good idea. This should be a job for editor formatters that you can configure, for example, via `.editorconfig`.

> But the formatter can't be right 100% of the time.

That's the point. It shouldn't be right all the time, specially because that's subjective. It should just be **consistent**. That's why it can't be configured.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+----
+module foo.bar.baz
+amends "bar.pkl"
+----
+
+==== [0005] Imports
+
+Imports should be sorted alphabetically. +
+Each import should be written in single line. +
+They should be grouped in 3 groups, in this order: absolute imports, relative imports, and project imports. +
```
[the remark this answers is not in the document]
stackoverflow: I'm fine with changing the order and making relative imports come last. As long as the rule is always the same for every file.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+
+== Detailed design
+
+The formatter won't allow any form of customization, similarly to https://pkg.go.dev/cmd/gofmt[gofmt].
+The idea is that there is a single canonical set of rules to format Pkl code. If people want to format their Pkl
+code differently, they can use other tools that allow customization.
+
+The standard library and https://github.com/apple/pkl-pantry[pkl-pantry] will be formatted using `pkl format`.
+
+Syntactically invalid modules cannot be formated. They will return an error when called via the CLI or the Java API.
```
HT154, line 35: s/formated/formatted/

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+and an ad-hoc formatter in the IntelliJ plugin, but that is only available in IntelliJ, has some quirks and bugs,
+and doesn't follow our style guide.
+
+A canonical formatter shipped with the language will allow users to format their Pkl code in an editor-independent way.
+It will also facilitate diffing Pkl code as formatting rules can be equalized.
+
+== Proposed Solution
+
+A new CLI command: `pkl format` will be introduced to the language to format Pkl files.
+A Java API will also be provided.
```
HT154, line 25: Would a gradle task make sense as well?

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+
+Some of these rules were taken from our https://pkl-lang.org/main/current/style-guide/index.html[style guide] verbatim,
+others had to be made more strict.
+
+Rules are presented here in order of more generic to less generic.
+Less generic rules (bigger numbers) overrule previously defined ones, if they conflict.
+
+==== [0001] Line Width
+
+Lines must be a maximum of 100 characters long. Exceptions are string literals, variable names, and doc comments.
```
HT154, line 96: 100 is really small! I've been enforcing 160 for projects with adjustable formatters recently and that's been a much happier medium.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+- After keywords, except function/property-like keywords (`trace`, `throw`, `module`, `this`, `super`, `outer`, `read`,
+`read?`, `read*`)
+- Before and after braces
+- Before and after an equals sign
+- Before and after an arrow (`->`)
+- Before a line or block comment start (`//`, `/*`)
+- Before a block comment end (`*/`)
+- Around infix operators
+- After a comma
+- After a colon
```
HT154, line 180: What about after a semicolon?

spices/SPICE-0014-canonical-formatter.adoc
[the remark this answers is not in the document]
HT154: Co-signed on changing the order here. This is also similar to what common go and python tooling do.

spices/SPICE-0014-canonical-formatter.adoc
HT154: Yeah this is a little weird to me, it seems like modifiers should be semantically grouped instead and groups should be specified in a fixed order. This rule as-is can produce odd recordings like `abstract local class` vs. `local open class`.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+local open class Foo extends Bar {}
+
+class ReaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaalyLongClassName
+  extends Bar {}
+----
+
+==== [0010] Object and class bodies
+
+Object and class bodies must have the opening `{` in the same line as the previous token. +
+A new line must follow the opening `{`. +
```
HT154, line 393: I'd really like to see some accommodation of "inline" object bodies as long as they're under the line length limit. This is in accordance with the style guide, and goes a long way to offset the syntactic "bulk" Pkl's object body syntax carries in comparison to languages like YAML.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  4
+}
+
+lineIsTooBig: Listing<Int> = new { 999999; 1000000; 1000001; 1000002; 1000003; 1000004; 1000005; 1000006 }
+----
+
+.output
+[source, pkl]
+----
+foo: Listing<Int> = new { 1; 2; 3; 4; 5; 6; 7 }
```
HT154, line 482: This is in violation of rule 0010, no?

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+}
+
+bar = new {
+  prop = 0
+}
+----
+
+==== [0011] Object members
+
+Object members must be defined in their own line. +
```
HT154, line 428: Similarly, it'd be nice to allow "inline" properties/elements (up to two), in accordance with the style guide.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  1000003
+  1000004
+  1000005
+  1000006
+}
+----
+
+==== [0013] Expressions
+
+Expression must either be defined in a single line on in separated lines. +
```
HT154, line 505: s/on/or/?

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+
+qux = new Listing {
+  1
+}
+----
+
+==== [0014] Object parameters
+
+Object parameters must start in the same line as the opening `{`, unless it exceeds the line limit. +
+A new line must follow the `->`.
```
HT154, line 552: Can an exception be made for single-property object bodies less than the line length limit?

spices/SPICE-0014-canonical-formatter.adoc
HT154: This rule seems like it'd massively bloat almost any marginally-complex "functional" code doing map/reduce/etc. operations

HT154 (commented):
Excited to see this coming together! I've got some broad feedback reconciling what's proposed here with my own muscle memory and the 80k+ lines of Pkl I maintain.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+.output
+[source, pkl]
+----
+module foo.bar.baz
+amends "bar.pkl"
+----
+
+==== [0005] Imports
+
+Imports should be sorted alphabetically. +
```
bioball, line 252: I think we should follow the style guide here and note that these should follow [natural sorting](https://en.wikipedia.org/wiki/Natural_sort_order)

spices/SPICE-0014-canonical-formatter.adoc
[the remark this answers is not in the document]
bioball: +1 on this; we're saying this right?

1. absolute URIs
2. project imports
3. relative imports

spices/SPICE-0014-canonical-formatter.adoc
bioball: Well, Pkl doesn't allow for multiple inheritance, so I think we're okay there

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  1000005
+  1000006
+}
+----
+
+==== [0013] Expressions
+
+Expression must either be defined in a single line on in separated lines. +
+If an expression spans multiple lines there must be a new line after the `=`. +
+If the expression is a `new` or amending expression it should start in the same line as the `=`. +
```
bioball, line 507: Can we follow the style guide here? It's simply:

1. There can optionally be a newline on the right-hand side of assignment
2. If there is a newline, the assignee must be indented one level deeper

That means this is okay:

```pkl
foo =
  new {
    ...
  }
```

And this is okay too:

```pkl
foo = 1
  + 2
  + 3
```

spices/SPICE-0014-canonical-formatter.adoc
bioball: FYI: this conflicts with 0010

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  else
+    1
+
+baz =
+  if (predicate)
+    0
+  else if (value)
+    1
+  else
+    2
```
bioball, line 613: I think we should allow for the then and else bodies being on the same line as the `if` and `else` keywords.

```pkl
  if (predicate) 0
  else if (value) 1
  else 2
```

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+bar = 1 + /* block comment */
+  1
+----
+
+.output
+[source, pkl]
+----
+// line comment
+foo =
+  1 // reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaly long comment
```
bioball, line 685: I'd actually move the line comment one line higher in this case

```pkl
// line comment
// reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaly long comment
foo = 1
```

bioball (commented):
Made a pass! Also: we should have rules around function parameters, and method call arguments.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: I'm fine with a different order, as long as it's easy to understand what would the result will be.
What would the groups and order be?
Locality (`external`) ->  changeability (`fixed`, `const`) -> visibility (`local`, `hidden`) -> extensibility (`abstract`, `open`)?
I'm open to suggestions.

madrob: Here's what exists today in `apple/pkl`
```
git grep -o -P '(external|fixed|const|local|hidden|abstract|open) (external|fixed|const|local|hidden|abstract|open) ' | cut -d: -f2 | sort | uniq -c
  12 abstract external
   3 abstract hidden
   2 abstract open
   2 const hidden
  18 const local
  24 external const
   2 hidden fixed
   1 local abstract
   8 local const
   2 local fixed
   2 local hidden
   5 local open
```

From that, I would guess `abstract external const local hidden fixed open`

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+
+function fun(x: Int?, b: Boolean) = if (b) /* return x */ x else x + bar[0]
+
+choices: "foo"|*"bar"|String
+----
+
+==== [0004] Module definitions
+
+Module, amend, and extend definitions must be written in a single line. +
+There must be no empty lines between a module and an amend/extend definition.
```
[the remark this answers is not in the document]
stackoverflow: Explained now in rule 0004.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+foo {
+  bar {
+    baz = "hi"
+  }
+}
+----
+
+==== [0008] Modifiers
+
+Modifiers must be written in the same line as what they are modifying, except if they go over the maximum line limit. +
```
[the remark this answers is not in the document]
stackoverflow: Example in rule 0004.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: Example is added to rule 0004.

spices/SPICE-0014-canonical-formatter.adoc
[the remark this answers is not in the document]
stackoverflow: Changed the order to: absolute -> project -> relative.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: We discussed this, but as @bioball mentioned, this is better done as a spotless integration.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: I'm fine with increasing this a bit, but we should be conservative here as this is not configurable. If it's too small it's just an annoyance. But too big would be a deal breaker for users. Something like 120 should be fine.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: I left too much unsaid here. This part of the rule is just for chains that already span multiple lines or are too big to fit in one. Property/method chain can for sure be written in a single line. Will clarify.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: I'm fine with allowing a single line. Same as the comment above merging 0011 and 0012.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: The problem I see is that natural sort is kind of "vibe" based. There's no canonical algorithm that everybody uses AFAIK. All the Java libs I found have their own idiosyncrasies on how they handle input. Even the one defined in naturalordersort.org has only 3 rules which I'm not sure is enough to handle all possible edge cases (like unicode, for example).
This is fine in the style guide because we are just telling people to manually sort their imports based on what they find natural. For the formatter, I'm not sure.
All that said, I'm not against using it here. We could implement our own flavor.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: I'm also fine with removing this rule completely and just respect whatever line width a user define for their code, never breaking lines due to a max limit.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: I'm fine with making this the rule, as long as we remove the other one.
My problem is having multiple ways to format the same thing.
We already allow single line and multi line expressions, which is a good compromise.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: You can't rearrange comments without changing their meaning in a generic way.
Sure it would work in the simple example above, but we can't generalize it.

spices/SPICE-0014-canonical-formatter.adoc
HT154: Yeah that seems reasonable to me. Is that in line with the goals of having a "canonical" formatter? It would imply that both the one-line and multi-line styles are valid and won't be changed by the formatter.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: Good catch!

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+  else if (value)
+    1
+  else
+    2
+----
+
+==== [0016] Let expressions
+
+The expression following the let must be on its own line, and it must be indented. +
+A let followed by another let must not be indented.
```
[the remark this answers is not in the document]
stackoverflow: Indeed. Missed that.

stackoverflow:
> we should have rules around function parameters, and method call arguments.
@bioball I added some rules for parameters and arguments in rule 0004. Do you think there's more missing?

spices/SPICE-0014-canonical-formatter.adoc
bioball: I find that adding a newline for some multiline expressions are more readable than others.

```pkl
foo =
  let (bar = getBar())
    new {
      res = bar
      res2 = bar
    }

multilineString =
  """
  Some very
  long
  multiline string
  """

myConditional =
  if (foo) someFooResult
  else if (bar) someBarResult
  else someBazResult
```

This is worse:

```pkl
foo = let (bar = getBar())
  new {
    res = bar
    res2 = bar
  }

multilineString = """
Some very
long
multiline string
"""

myConditional = if (foo) someFooResult
else if (bar) someBarResult
else someBazResult
```

Same with multiline string literals, multiline if/else expressions

bioball: However, other multiline expressions are perfectly fine to start on the same line, and so far, we have been writing it that way. `new` expressions are a good example of that.

I think it'd be good to allow for this:

```
foo = new {
  bar = foo
}
```

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,780 @@
...
+bar =
+  (if (true) 10 else -10) +
+    100 *
+    someFuntion(1, 2, "foo")
+
+baz =
+  1 +
+    3 *
+    5 -
+    9
```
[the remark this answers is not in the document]
bioball: Not sure about most? I think it's generally pretty split.

Here's what I see from a quick test of different formatters

* google-java-format: after the newline
* ktfmt: before the newline
* black (python): after the newline
* prettier (js/ts): before the newline
* gofmt: before the newline

Actually, I discovered that many of these formatters have different rules for different binary operators.
For example, ktfmt formats the elvis operator like this:

```ktfmt
    var foo =
      bar
        ?: orReaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaallyLongFooValue
        ?: orReaaaaaaaaaaaaaaaaaaaaaaaaaaaaallyLongBarValue
```

But the plus operator like this (required because kotlin does not accept `+` after the newline)

```kt
    var foo =
      bar +
        orReaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaallyLongFooValue +
        orReaaaaaaaaaaaaaaaaaaaaaaaaaaaaallyLongBarValue
```

I noticed the same thing in Prettier, with `+` and `|>` being formatted differently.

The pipe operator feels especially hard to read if the operator is placed before the newline:

```pkl
foo =
  bar |>
    orReaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaallyLongFooValue |>
    orReaaaaaaaaaaaaaaaaaaaaaaaaaaaaallyLongBarValue
```

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: Yes, that's a good question.
An issue we might have is that if the formatter is too strict people might be so annoyed they won't use it.
I'd be OK with forcing multi line style for all members, but I imagine a lot of users would hate that.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: > Maybe we can have a hard set of rules for what should come after the newline

That seems reasonable.
My idea is that `new` and amends should start on the same line. Everything else should indent.

spices/SPICE-0014-canonical-formatter.adoc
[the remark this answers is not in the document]
stackoverflow: I think that's something we should fine tune when we have a working version of the formatter to see how readable it is.
Anyway, as long as we have a single way to format an operator (even if it differs per operator) it should be fine.

spices/SPICE-0014-canonical-formatter.adoc
bioball: Yeah; we can play around with this and see what works.

The biggest thing I'd like to solve for is to sort numbers in order.

```pkl
import "cluster11.pkl"
import "cluster2.pkl"
```

Should become:

```pkl
import "cluster2.pkl"
import "cluster11.pkl"
```

spices/SPICE-0014-canonical-formatter.adoc
HT154: It may be worth considering making line length the one configurable option. Formatters in the Python space (black, ruff) take this approach. I think 120 characters (fixed or as a default) would be pretty reasonable though.

spices/SPICE-0014-canonical-formatter.adoc
bioball: I think I'm okay with a simple rule like: if it fits within the column limit, make it single line. Otherwise, make it multiline.

I also think we need to play around with this with real-world code to see what it looks like in practice.

spices/SPICE-0014-canonical-formatter.adoc
bioball: This is totally subjective, but, 100 chars has been a totally reasonable column limit for me; that's what our own Pkl code adheres to, and that also happens to be how most of our code is formatted. 120 is probably fine too.

I also really like the idea of a max column width. Without this, it can lead to quite hard to read code. Also, the max column width can be used to guide how formatting works (e.g. when to break single-line object bodies into multi-line).

I'm not inclined to make this configurable; it would be good to provide a consistent rule for all Pkl code.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,906 @@
...
+  1 // reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaly long comment
+
+bar =
+  1 + /* block comment */
+    1
+----
+
+==== [0019] Type aliases
+
+Type aliases must be written in a single line if they fit the line limit.
```
gordonbondon, line 815: How will this format long typealiases `typealias String "a"|"b" very long...|"xy123"`, each option on new line, or try to fit options on one line put to line limit?

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,906 @@
...
+[source, pkl]
+----
+module foo.bar.baz
+amends "bar.pkl"
+----
+
+==== [0006] Imports
+
+Imports must be sorted alphabetically. +
+They must be grouped in 3 groups, in this order: absolute imports, project imports, and relative imports. +
```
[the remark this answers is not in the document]
HT154: Unlike with go, all stdlib imports will group together by nature of sharing a common prefix. I wouldn't be opposed to special handling, but also I don't think it's critically important here.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,906 @@
...
+
+N/A
+
+== Future directions
+
+More rules can be added (though we should be wary of altering existing rules). And with parse time variable resolution
+we could not only format but simplify code. +
+Example of things that could be simplified in the future:
+
+- Remove unused imports
```
HT154, line 891: I think, if a formatter/linter ever does this it also needs to have some manner of automatic import addition as well, a la `goimports`. As a proponent of format-on-save, I'd find an import removed by a save mid-refactor to be more than mildly annoying. For project dependencies this seems reasonable enough to accomplish.

spices/SPICE-0014-canonical-formatter.adoc
bioball: I think of trailing line comments as always targeting the whole line preceding it. IMO: moving it one line higher always means the same thing.

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: Indeed. That's why I didn't add this to the rules. The formatter should not change your original code, just format it.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,906 @@
...
+const local function looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogName(x: Int, y) = x + y
+
+const local function reallyLooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogName(x: Int, y) = x + y
+----
+
+.lineBreakOutput
+[source, pkl]
+----
+module
+  reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaly
```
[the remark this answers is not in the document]
stackoverflow: `<identifier>` is considered a single identifier and it cannot be broken (rule 0001).

spices/SPICE-0014-canonical-formatter.adoc
stackoverflow: Good one! I added type examples to rule 0004.

spices/SPICE-0014-canonical-formatter.adoc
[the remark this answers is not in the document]
gordonbondon: With current rule imports togather with stdlib will look like this:
```pkl
import "https://example.com/baz.pkl"
import "package://example.com/myPackage@1.0.0#/Qux.pkl"
import "pkl:semver"
import "pkl:json"

import "@bar/Bar.pkl"
import "@foo/Foo.pkl" as foo
```

In case of separate group for stdlib it will be:
```pkl
import "pkl:semver"
import "pkl:json"

import "https://example.com/baz.pkl"
import "package://example.com/myPackage@1.0.0#/Qux.pkl"

import "@bar/Bar.pkl"
import "@foo/Foo.pkl" as foo
```

Which IMO would be nicer.

spices/SPICE-0014-canonical-formatter.adoc
```diff
@@ -0,0 +1,925 @@
...
+import "@bar/Bar.pkl"
+import "@foo/Foo.pkl" as foo
+
+import* "**.pkl"
+import "..."
+----
+
+==== [0007] Module and class members
+
+Module and class members must be separated by exactly one empty line.
```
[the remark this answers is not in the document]
stackoverflow: Changed this to be `at max one empty line` so the original code separation is preserved.
One exception is, if there was no line separation in the original code, one empty line will be added:
```pkl
class Foo { foo = 1; bar = 2 }
```
will become
```pkl
class Foo {
  foo = 1

  bar = 2
}
```

PR #17, opened 2025-05-06
SPICE-0016: HTTP Rewrites

bioball:


spices/SPICE-0016-http-rewrites.adoc
```diff
@@ -0,0 +1,310 @@
...
+----
+Pkl Error
+An I/O error occurred when making request GET /foo:
+Connection reset (request was rewritten: https://foo.com -> https://bar.com)
+----
+
+[[configuration-api]]
+=== Configuration API
+
+HTTP rewrites can be configured via the standard library, in Java as an API, as a server message API, and as a Gradle API.
```
HT154, line 84: And via the CLI? I guess technically not an API but seems odd to make that distinction here.

Also, I think it might help to clarify how these different interfaces override each other. Looking at the implementation, it appears that any configure rewrites in a project overwrites all rewrites configured in settings.pkl. Likewise, any rewrites configured on the CLI override all rewrites configured via PklProject or settings.pkl.

spices/SPICE-0016-http-rewrites.adoc
```diff
@@ -0,0 +1,310 @@
...
+  --http-rewrite https://foo.com=https://bar.com \
+  myModule.pkl
+----
+
+This flag will be available to commands that use common CLI options (for example, `eval`, `test`, `project`).
+
+=== Security checks
+
+All outbound requests will be checked against the `--allowed-modules` and `--allowed-resources` CLI flags, and their respective settings in other contexts.
+This is also the case with URL rewriting.
```
HT154, line 206: Are the pre-rewrite URIs checked? Post-rewrite? Both?

spices/SPICE-0016-http-rewrites.adoc
```diff
@@ -0,0 +1,310 @@
...
+
+=== Error modes
+
+==== Invalid scheme
+
+An HTTP target URL must start with `http://` or `https://`.
```
HT154, line 212: I wonder if it makes sense to prevent rewriting HTTPS URIs to HTTP URIs for security URIs. This seems like a potential "downgrade" sort of attack surface.

bioball: I've considered this, but I don't think this is necessary.

If users want to disable HTTP schemes, the correct path for that is to use the `--allowed-modules` and `--allowed-resources` flags.

spices/SPICE-0016-http-rewrites.adoc
```diff
@@ -0,0 +1,310 @@
...
+==== Standard Library
+
+A new property will be added to class `pkl.EvaluatorSettings.Http`.
+
+.EvaluatorSettings.pkl
+[source,pkl]
+----
+class Http {
+  @Since { version = "0.29.0" }
+  rewrites: Mapping<Uri, Uri(startsWith("http://") || startsWith("https://"))>?
```
stackoverflow, line 98: Why only constrain the value to start with http(s) and not the key? This class being called `Http` kinds of implies that you can only rewrite http URLs.

stackoverflow (approved):

spices/SPICE-0016-http-rewrites.adoc
```diff
@@ -0,0 +1,310 @@
...
+    ["https://pkg.pkl-lang.org/"] = "https://my.internal.mirror/pkg-pkl-lang/"
+    ["https://github.com/"] = "https://my.internal.mirror/github-releases/"
+  }
+}
+----
+
+With this setup, Pkl will make HTTP calls to `https://my.internal.mirror` when downloading Pkl packages.
+In practice, this would be a server that fetches those assets from the underlying locations and responds with the same bytes.
+
+== Detailed design
```
stackoverflow, line 62: I miss the explanation for what happens when 2+ keys match a URL.
For example, if I have:
```pkl
["https://foo.bar/baz"] = "<rewrite1>"
["https://foo.bar/"] = "<rewrite2>"
```
Which one is going to win when we fetch `foo.bar/baz/qux`? I'd expect the more specific (rewrite1) to win.

bioball: Ah, thanks. In this case, the longest prefix should win.

Also, I think there's something else to consider here that I didn't initially think about: in a URI, the domain and the scheme are both case-insensitive.

These two URIs are equivalent:

```
HTTPS://EXAMPLE.COM
https://example.com
```

So, this is an ambiguous definition:

```
rewrites {
  ["HTTPS://EXAMPLE.COM"] = "<rewrite1>"
  ["https://example.com"] = "<rewrite2>"
}
```

bioball: Moreover, you should not be able to bypass a rewrite rule by changing the case in your Pkl code.

A rewrite rule targeting `https://example.com/` should be used to replace this import:

```
import "HTTPS://EXAMPLE.COM/foo.pkl"
```

spices/SPICE-0016-http-rewrites.adoc
bioball: Yeah; good point. The key should also have this constraint.

spices/SPICE-0016-http-rewrites.adoc
bioball: Good question. I think both.

spices/SPICE-0016-http-rewrites.adoc
stackoverflow: We should also consider that case doesn't matter so `"HTTPS://..."` is also valid.

spices/SPICE-0016-http-rewrites.adoc
stackoverflow: I think only outbound should be checked.
If you only allow `https` and there's an `http` import/read in the code that is rewritten to `https`. That should be valid, as, in the end, you are not calling an insecure endpoint.
This would allow people to use libraries that import/read resources that they forbid by rewriting them.

PR #21, opened 2025-08-12
SPICE-0019: Trailing commas

HT154:


spices/SPICE-0019-trailing-commas.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+// after formatting
+x = myMethod(1, 2, 3) // <1>
+y = myMethod(
+  1, 2, 3, // <2>
+)
+z = myMethod(
+  1,
+  2,
+  3, // <3>
```
stackoverflow, line 69: The formatter doesn't respect newlines in method calls so both `x`, `y`, and `z` will be formatted as `myMethod(1, 2, 3)`.
Also `y` can never happen. If the method call has to break, every element will be put in a new line (like `z`).
I think it's fine to leave the examples as they are but adding a comment that they are illustrative, and not how the formatter would format these constructs in real code.

spices/SPICE-0019-trailing-commas.adoc
```diff
@@ -0,0 +1,85 @@
...
+z = myMethod(
+  <expr>,
+  <expr>,
+  <expr>
+)
+
+// after formatting
+x = myMethod(<expr>, <expr>, <expr>) // <1>
+y = myMethod(
+  <expr>, <expr>, <expr>, // <2>
```
bioball, line 65: This looks like weird formatting; not sure if our formatter would produce this?

HT154: It can produce something similar for `Map()` invocations with line-paired arguments. Per the note below, this is just to illustrate the rule.

PR #24, opened 2025-09-23
SPICE-0022: Custom HTTP Headers

kyokuping:
This proposes a new option for HTTP headers .

Reference implementation: https://github.com/apple/pkl/pull/1196

### DIFF
--- a/spices/SPICE-0022-http-headers.adoc
+++ b/spices/SPICE-0022-http-headers.adoc
@@ -1,0 +1,92 @@
+= Custom HTTP Headers
+
+* Proposal: link:./SPICE-0022-http-headers.adoc[SPICE-0022]
+* Author: link:https://github.com/kyokuping[kyokuping]
+* Status: Proposed
+* Category: Language, Standard Library, Tooling
+
+== Introduction
+
+This SPICE proposes a new evaluator option to allow adding custom HTTP headers to outbound requests, with the ability to target specific URL prefixes.
+
+== Motivation
+
+When Pkl interacts with HTTP resources, it may need to provide authentication tokens or other custom headers to access them. For example, a Pkl module might be hosted in a private repository that requires an `Authorization` header for access.
+
+Currently, there is no way to add custom headers to Pkl's HTTP requests. This limits Pkl's ability to interact with a wide range of HTTP-based resources.
+
+== Proposed Solution
+
+A new HTTP setting will be added, called `headers`. This setting allows users to specify a list of headers to be added to outbound HTTP requests. To avoid leaking credentials, headers can be configured on a per-URL-prefix basis.
+
+For example, to provide an authentication token for a specific host, the configuration would look like this:
+
+.~/.pkl/settings.pkl
+[source,pkl]
+----
+amends "pkl:settings"
+
+http {
+  headers {
+    ["https://my.private.repo/"] {
+      Pair("Authorization", "Bearer my-secret-token")
+    }
+  }
+}
+----
+
+This will add the `Authorization` header to all requests sent to `https://my.private.repo/` and its sub-paths.
+
+== Detailed design
+
+The `pkl.EvaluatorSettings.Http` class will be extended with a new `headers` property.
+
+.pkl:EvaluatorSettings
+[source,pkl]
+----
+class Http {
+  // ... existing properties
+
+  /// HTTP headers to add to outbound requests targeting specified URLs.
+  headers: Mapping<HttpRewrite, Listing<Pair<String, String>>>?
+}
+----
+
+The `HttpRewrite` typealias from `SPICE-0016-http-rewrites` will be reused to define the URL prefixes.
+
+When multiple prefixes match a request URL, the longest prefix will be chosen, ensuring that the most specific rule is applied.
+
+A new CLI flag, `--http-header`, will be introduced to allow specifying headers from the command line. The flag will accept key-value pairs in the format `<url-prefix>=<header-name>:<header-value>[,<header-name>:<header-value>...]`.
+
+Example:
+[source,shell]
+----
+pkl eval \
+  --http-header "https://my.private.repo/:Authorization=Bearer my-secret-token" \
+  myModule.pkl
+----
+
+The HTTP client will be modified to check for matching header configurations for each outgoing request and add the corresponding headers.
+
+== Compatibility
+
+This change is strictly backward-compatible. Existing Pkl programs will continue to work as-is.
+
+== Future directions
+
+Future enhancements could include support for more complex header manipulation, such as removing default headers or adding headers based on regular expression matching on the URL.
+
+== Alternatives considered
+
+=== Global HTTP headers
+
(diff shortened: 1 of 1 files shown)


spices/SPICE-0022-http-headers.adoc
```diff
@@ -0,0 +1,92 @@
...
+----
+amends "pkl:settings"
+
+http {
+  headers {
+    ["https://my.private.repo/"] {
+      Pair("Authorization", "Bearer my-secret-token")
+    }
+  }
+}
```
bioball, line 35: Not sure about using `Pair` here. I think this should just be:

```pkl
headers: Mapping<String, Mapping<String, String>>
```

To address @HT154 's concern from https://github.com/apple/pkl/pull/1196#pullrequestreview-3171835904

> Although rare, it is valid to send a specific header more than once in a request. When there are headers with the same name their order may also matter.

This is quite rare indeed; I don't know if it's worth supporting. In most cases, a comma separator semantically represents "multiple values" (e.g. `Accept: text/plain,application/json`). Seems like the one case where this is not true is the `Set-Cookie` header (see https://stackoverflow.com/a/6375214).

I don't know if Pkl users will ever need the `Set-Cookie` header; I think just `Mapping<String, String>` very likely will address all user needs.

If we _did_ want to support multiple of the same header, I think the type should instead be:

```pkl
headers: Mapping<String, Mapping<String, *Listing<String>|String>>>
```

Where `Listing` means: repeat this header multiple times.

spices/SPICE-0022-http-headers.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+== Future directions
+
+Future enhancements could include support for more complex header manipulation, such as removing default headers or adding headers based on regular expression matching on the URL.
+
+== Alternatives considered
+
+=== Global HTTP headers
+
+The initial proposal involved a simpler approach of having a single set of custom headers that would be applied to all outbound HTTP requests. This was rejected due to security concerns. Sending the same set of headers, which might include authentication tokens, to all hosts is a significant security risk. The per-URL-prefix approach provides a more secure way to manage custom headers.
```
bioball, line 83: It's valid to want headers to apply to all requests; for example, you might want all of Pkl's requests to use your own user-agent header.

We can perhaps have `*` be a special case that represents all hosts?

spices/SPICE-0022-http-headers.adoc
bioball: Some other notes:

* Header names are case-insensitive. We should enforce that you cannot repeat the same header twice with different casings.
* Some headers make no sense to allow users to set (e.g. `Host`, `Content-Length`). I think we should also enforce that these headers _cannot_ be set.

We can validate this in Pkl. Perhaps we can require that header names are lower-case to begin with:

```pkl
typealias ReservedHeaderName =
  "host"
  | "content-length"
  | "// etc"

typealias HeaderName = String(this == toLowerCase(), !(this is ReservedHeaderName))

headers: Mapping<String, Mapping<HeaderName, String>>
```

spices/SPICE-0022-http-headers.adoc
```diff
@@ -0,0 +1,92 @@
...
+http {
+  headers {
+    ["https://my.private.repo/"] {
+      Pair("Authorization", "Bearer my-secret-token")
+    }
+  }
+}
+----
+
+This will add the `Authorization` header to all requests sent to `https://my.private.repo/` and its sub-paths.
```
bioball, line 38: Many auth methods require some sort of extra logic to form a token (for example, some process to fetch a short-lived cookie).

It'd be good to think about how this should work in Pkl.

Possibly, users can use a [custom resource reader](https://pkl-lang.org/main/current/language-reference/index.html#extending-resource-readers):

```pkl
headers {
  ["https://my.private.repo/"] {
    ["Cookie"] = read("auth:get-my-cookie")
  }
}
```

One issue here is that custom resource readers _are not available_ when evaluating `~/.pkl/settings.pkl` and `PklProject`. This is a chicken-and-egg problem; custom readers can possibly be configured here in the first place.

This isn't necessarily blocking this feature, but something to think about as we design t his.

spices/SPICE-0022-http-headers.adoc
```diff
@@ -0,0 +1,92 @@
...
+The `pkl.EvaluatorSettings.Http` class will be extended with a new `headers` property.
+
+.pkl:EvaluatorSettings
+[source,pkl]
+----
+class Http {
+  // ... existing properties
+
+  /// HTTP headers to add to outbound requests targeting specified URLs.
+  headers: Mapping<HttpRewrite, Listing<Pair<String, String>>>?
```
bioball, line 51: We should probably rename `HttpRewrite` to `HttpPrefix` or something like that.

bioball (commented):
Thanks for writing this! Some initial feedback.

spices/SPICE-0022-http-headers.adoc
bioball: Yeah, and with a deprecation:

```pkl
@Deprecated { since = "x.x.x"; replaceWith = "HttpPrefix" }
typealias HttpRewrite = HttpPrefix
```

spices/SPICE-0022-http-headers.adoc
bioball: That's possible, although, it differs from how HTTP rewrites work. Theoretically, we can change HTTP rewrites to follow this logic too, but it would be a breaking change:

  - We determine precedence by _longest prefix_. This makes sense right now because these prefixes are verbatim. However, this doesn't hold if we use wildcards (the longest wildcard is not necessarily the most specific).
  - Wildcards are also URL characters, so existing valid prefixes would then be re-interpreted as a wildcard.

If we accept patterns, glob patterns would probably be better than regexes (glob patterns are designed to work well with paths, `.` is not a special character).

Perhaps we can have HTTP headers and HTTP rewrites just have different rules too (one is a verbatim match, one is a wildcard match), because these address different problems in the first place and are different solutions (mirroring vs. authentication).

spices/SPICE-0022-http-headers.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+The HTTP client will be modified to check for matching header configurations for each outgoing request and add the corresponding headers.
+
+=== The `URLPattern` typealias
+
+The `URLPattern` typealias defines glob patterns that are used to match the URLs of HTTP requests and determine which HTTP headers to apply.
+
+To prevent accidental matches, the pattern should end with a slash (`/`) or a wildcard (`*`).
+
+If several patterns match a given request URL, the one with the highest specificity is selected. The specificity of a pattern is evaluated using the following process:
```
kyokuping: After implementing this, I realized that people will expect to have all matching patterns to be selected, instead of a single one. What do you think? @bioball @HT154

kyokuping: I'll proceed with selecting all matching patterns for now.

bioball: Yeah, I think "all matches apply" makes sense too! Apologies for the delay; was on pat leave.

PR #29, opened 2025-12-23
SPICE-0026: Power Assertions

bioball:
Reference implementation: https://github.com/apple/pkl/pull/1384


spices/SPICE-0026-power-assertions.adoc
```diff
@@ -0,0 +1,307 @@
...
+== Future directions
+
+=== Assertions as an API
+
+Possibly, we can provide an in-language assertion API.
+For example:
+
+[source,pkl]
+----
+assert(1 == 2)
```
stackoverflow: I know this a future direction, but would assert short circuit if it fails, like throw?
If so it has the same problem as throw, which is, it doesn't compose.
If we can solve this, we'd be able to replace throw as the way to show custom messages.

PR #31, opened 2026-04-08
SPICE-0028: Line Continuation for Multi-line Strings

HT154:


spices/SPICE-0028-multiline-string-continuation.adoc
```diff
@@ -0,0 +1,66 @@
...
+== Detailed design
+
+Support line continuations requires some minor changes in several places:
+
+* `pkl-parser`
+** Add a `Token` and support for it in `Lexer`.
+** Add support for parsing continuations in `Parser` and `GenericParser`.
+* `pkl-core` - Mark line continuations as a string escape in `SyntaxHighlighter`.
+* `pkl-formatter` - Ensure `Builder` formats continuations correctly.
+* `pkl-intellij` - Add support for line continuations in the grammar.
```
bioball: Also, pkl-lsp and pkl-intellij should both be able to constraint-check these correctly.

There should be a squiggly here:

```pkl
foo: String(startsWith("Hi bub")) = """
  Hi \
  bob
  """
```

spices/SPICE-0028-multiline-string-continuation.adoc
```diff
@@ -0,0 +1,66 @@
...
+* `tree-sitter-pkl` - Add support for line continuations in the grammar.
+** Changes must be adopted in `pkl-lsp`, `pkl-vscode`, and `pkl-neovim`
+* `highlightjs-pkl` - Add support for line continuations in the grammar.
```
bioball, line 121: This might be fine as-is; highlightjs-pkl is regex-based and doesn't actually understand the parse tree. As long as it's highlighted as part of the string, it should be okay.

Same with our textmate grammar.

spices/SPICE-0028-multiline-string-continuation.adoc
```diff
@@ -0,0 +1,66 @@
...
+* Implemented in: Pkl 0.32
+* Category: Language, Tooling
+
+== Introduction
+
+Grammar and parser updates to support `\` as a line continuation character in multi-line string literals.
+
+== Motivation
+
+Pkl aims to align its syntax for string literal with Swift's syntax.
```
bioball: I'd say that it's not necessarily a goal to align with Swift; we're okay with some differences here and there. But, Swift having prior art here is a strong motivator.

stackoverflow: I'd go even further and say that this is a non-goal. The syntax was inspired by Swift, but we don't have to follow it. Specially because Swift is constantly evolving, so I don't think chasing its syntax is a good idea.

spices/SPICE-0028-multiline-string-continuation.adoc
HT154: Looks like Swift's hljs grammar does specifically handle this: https://github.com/highlightjs/highlight.js/blob/5697ae5187746c24732e62cd625f3f83004a44ce/src/languages/swift.js#L142

spices/SPICE-0028-multiline-string-continuation.adoc
HT154: Updated this to note that matching Swift is a non-goal

spices/SPICE-0028-multiline-string-continuation.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+In literals with custom delimiters, the `\` must be followed by the same number of `#` used by the literal's delimiters.
+[source,pkl]
+----
+local multiLine =
+  #"""
+  Hello, \#
+  World!
+  """#
+----
```
bioball, line 101: Should note: it's an error to have whitespace following `\`. It's also an error to use `\` in a single-line string literal.

spices/SPICE-0028-multiline-string-continuation.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+It is invalid for any whitespace other than a newline (`\n`) to follow the `\`.
+Line continuation is only applicable to multi-line string literals; using one in a single-line string literal is an error.
```
bioball: Should add some details about error modes here. Some questions:

1. What happens if there's non-newline whitespace after the `\` character? This is an error, right?
2. What happens if you use this in a single-line string?

bioball (approved):
In general, this feature makes sense to me, and I'm on board with adding this to Pkl.

PR #32, opened 2026-07-14
SPICE-0029: Self Type

HT154:


spices/SPICE-0029-self-type.adoc
```diff
@@ -0,0 +1,77 @@
...
+Only a couple Pkl API changes are directly required as part of adding the `this` type:
+
+* Add `pkl.reflect#ThisType`
+* Add `pkl.reflect#thisType`
+
+A couple more changes will be made to adopt the `this` type in existing APIs:
+
+* `Any.getClass()` will return `Class<this>` instead of just `Class`
+** This allows removing special-cased workarounds from https://github.com/apple/pkl-intellij/pull/202[pkl-intellij] and https://github.com/apple/pkl-lsp/pull/188[pkl-lsp].
+* `pkl.ref#Domain.renderReference()` now accepts a `Reference<this, Any>` argument instead of `Reference<Domain, Any>`
```
bioball: Also `Any.ifNonNull()`:

```pkl
external function ifNonNull<Result>(transform: (this) -> Result): Result?
```

Not perfect; but much better than using `NonNull`

Maybe in the future we can introduce a type param for `NonNull` so this can become `NonNull<this>`

spices/SPICE-0029-self-type.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+== Detailed design
+
+Pkl's parsers will be updated to recognize `this` in type positions.
+Similarly, pkl-intellij and pkl-lsp (incl. tree-sitter-pkl) will be updated to parse and resolve `this` as a type.
+
+=== Usage Restrictions
+
+Semantics of self types inside type alias bodies are not easily definable, especially when the enclosing module is not final.
+Usage of both `this` in such contexts violates the existing restriction on accessing only `const` values.
```
bioball: I think mentioning `const` here is a little confusing. Maybe:

```suggestion
A self type is not compatible with typealiases.
A typealias is meant be "static"; an extending module does not change the meaning of a typealias (see https://github.com/apple/pkl-evolution/blob/main/spices/SPICE-0007-const-checks-in-typealiases.adoc[SPICE-0007]).
On the other hand, the `this` type always refers to the type of the extending module.
Because of this, the `this` type cannot be used inside typealias bodies.
```

spices/SPICE-0029-self-type.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+As such, the `this` type is forbidden for use inside typealias bodies.
+
+The `module` type has similar issues when used inside class bodies and typealias.
+A new deprecation warning for use of the `module` type in these contexts will be introduced and will become an evaluation error in a subsequent release.
+In most scenarios, the `module` type should be replaced with a self-import.
```
bioball: Some thoughts:

* Can you add something to the intro to the line of "this SPICE also refines the module type"?
* I think this needs its own section; what does "deprecation warning" mean? What are examples of invalid usage of the `module` type?
* How long is the deprecation cycle?
* Should we provide a migrator? Quickfix?

spices/SPICE-0029-self-type.adoc
```diff
@@ -0,0 +1,77 @@
...
+The `module` type also means different things at the module level vs. inside a class or typealias body, which can be confusing.
+
+== Proposed Solution
+
+A new type `this` will be added that serves as a self type in both module and class contexts.
+
+== Detailed design
+
+Pkl's parsers will be updated to recognize `this` in type positions.
+Similarly, pkl-intellij and pkl-lsp (incl. tree-sitter-pkl) will be updated to parse and resolve `this` as a type.
```
bioball, line 27: Can we include sections around:

* Prior art and naming

  Swift, Rust, TypeScript, Python, and Zig each have the concept of "self type":

  Swift: `Self`
  Rust: `Self`
  TypeScript: `this`
  Python: `Self`
  Zig: `const Self = @This()`

  Swift/Rust/Python have a `self/Self` distinction: `self` talks about the instance, and `Self` talks about the type.

  However, `this` feels more appropriate for Pkl for both the type and the instance because we already have this parity with `module` (it talks about both the instance and the type). Also, this is nice because it's a non-breaking change.

* Type checking semantics (what is well typed, what becomes type errors, etc)

HT154: I think I covered the semantics in enough depth, but let me know if there's anything I need to spend more detail on!

spices/SPICE-0029-self-type.adoc
```diff
@@ -0,0 +1,224 @@
...
+
+* Its casing aligns with Pkl's existing keyword types (`module`, `unknown`, `nothing`).
+* Using `this` avoids allocating a new keyword, which is a breaking change for code using that name as an identifier.
+** Choosing `self` is particularly problematic as it's common in "scope capture" scenarios often written as `local self = this`.
+* Using `this` to mean a type or a value depending on context matches the existing behavior for `module`.
+
+=== Semantics
+
+The `this` type is an alias for the class of the current receiver (`this` value).
+For open classes, `this` admits values whose class is a subclass of the receiver.
```
bioball, line 51:
```suggestion
For open or abstract classes, `this` is the specific subclass of that class
```

bioball (commented):
There's a fair bit left to define about what the `this` type means; e.g.

* in object body param (`foo { bar: this -> }`) (should be the outer enclosing object, not the inner object!)
* inside object inside class body (`class Foo { obj { local bar: this = ... } }`)
* in let expression (`let (foo: this) ...`)
* in for generator (`for (foo: this in bar)`)
* in annotation `@this { foo = ... }`
* explicit new expr (`new this {}`)

Do you mind going through the grammar and detailing each of these scenarios?

Also: maybe we can structure this SPICE by talking about what this means in different parts of the grammar.
Then, note about resolving to the visibly outer scope can be a sub-bullet point of type test/type cast expressions I think.

PR #34, opened 2026-08-14
SPICE-0031: OS Base Directories

bioball:
Implementation here: https://github.com/apple/pkl/pull/1809

### DIFF
--- a/spices/SPICE-0031-os-base-directories.adoc
+++ b/spices/SPICE-0031-os-base-directories.adoc
@@ -1,0 +1,146 @@
+= OS Base Directories
+
+* Proposal: xref:SPICE-0031-os-base-directories.adoc[SPICE-0031]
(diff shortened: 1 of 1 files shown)


spices/SPICE-0031-os-base-directories.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+Documented default paths (in the CLI reference, Gradle plugin docs, and the `pkl.settings` module) are updated accordingly.
+
+== Future directions
+
+=== Removing the legacy fallback
+
+A future version of Pkl could stop falling back to `~/.pkl/settings.pkl` and `~/.pkl/cacerts/`, once enough time has passed for users to migrate.
+
+== Alternatives considered
```
HT154, line 127: Should we also document the alternative of using native dirs on macOS (under ~/Library)?

PR #36, opened 2026-09-16
SPICE-0033: URL support in the standard library

stackoverflow:


spices/SPICE-0033-url-library.adoc
```diff
@@ -0,0 +1,373 @@
...
+  ///
+  /// A URL has an authority exactly when it has a [host].
+  /// This is [userInfo], [host] and [port] joined back together, encoded as in [toString()].
+  external fixed authority: String?
+
+  /// The percent-decoded segments of [path].
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("https://example.com/a/b%20c").segments == List("a", "b c")
```
bioball: Shouldn't this be `List("", "a", "b c")`?

Otherwise, we can't differentiate between these two:
* `foo:bar/baz/qux`
* `foo:/bar/baz/qux`

spices/SPICE-0033-url-library.adoc
```diff
@@ -0,0 +1,373 @@
...
+  /// ```
+  external fixed segments: List<String> // <3>
```
bioball:
```suggestion
  external fixed pathSegments: List<String> // <3>
```

spices/SPICE-0033-url-library.adoc
```diff
@@ -0,0 +1,373 @@
...
+  /// A name that repeats holds the values of all its occurrences, in the order they are written.
+  /// A parameter that states no `=` holds no value at all, and is left out; this is not the same as
+  /// a bare `=`, which gives it the empty string.
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("https://example.com/?q=a+b&q=c").queryParameters["q"] == new Listing { "a b"; "c" }
+  /// net.Url("https://example.com/").queryParameters.isEmpty
+  /// ```
+  external fixed queryParameters: Mapping<String, Listing<String>> // <4>
```
bioball: Inconsistency: `segments` gives you collection types, but this gives you an object type.

I think collection types makes sense, since these properties aren't amendable (same treatment as `reflect.Class.allProperties`, etc).

```suggestion
  external fixed queryParameters: Map<String, List<String>> // <4>
```

spices/SPICE-0033-url-library.adoc
```diff
@@ -0,0 +1,373 @@
...
+  ///
+  /// Throws if [ref] is a string that cannot be parsed, or if this URL is itself a relative
+  /// reference.
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("https://example.com/a/b").resolve("../c").toString() == "https://example.com/c"
+  /// net.Url("https://example.com/a/b").resolve("#top").toString() == "https://example.com/a/b#top"
+  /// ```
+  external function resolve(ref: String | Url): Url // <5>
```
bioball: Going back on our conversation from earlier: I actually think it would be good to offer a `normalize()` method; because it behaves differently than `resolve(".")`.

Given RFC-3986 semantics, I'd expect these to be true:

```pkl
Url("foo:/bar/baz").normalize().toString() == "foo:/bar/baz"

Url("foo:/bar/baz").resolve(".").toString() == "foo:/bar/"
```

spices/SPICE-0033-url-library.adoc
```diff
@@ -0,0 +1,373 @@
...
+/// ```
+external const function parseQuery(query: String): Mapping<String, Listing<String>>
+
+external local const function isValidPercentEncoding(value: String): Boolean // <3>
+
+external local const function isValidScheme(value: String): Boolean // <3>
+
+external local const function isValidHost(value: String): Boolean // <3>
+
+external local const function isValidPath(value: String, hasAuthority: Boolean): Boolean // <3>
```
bioball: Any reason to mark these `local`?

Also, maybe move these inside the `Url` class, or change the name to include "Url" (e.g. `isValidUrlPath`)

spices/SPICE-0033-url-library.adoc
```diff
@@ -0,0 +1,373 @@
...
+if there is one.
+
+==== Message format
+
+When serialized, `net.Url` will be encoded as a `Typed` object (see https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html#non-primitives).
+
+The client in the host language will turn the URL into the target language via their respective codec/mapper/unmarshaller.
+
+When decoding, the language has a choice whether construct the URL either component-wise (using a constructor
+and using `scheme`, `host`, etc. individually), or by constructing the URL using the serialized string form.
```
bioball: The pkl-binary encoding for `Url` would only encode its non-hidden, non-external properties, none of which includes the string form.

So, the decoder would need to construct one component-wise.

spices/SPICE-0033-url-library.adoc
stackoverflow: There are many ways to go about this.
I copied the design here from [Dart](https://api.dart.dev/dart-core/Uri/pathSegments.html) and [furl](https://github.com/gruns/furl).
One can always check the path to see if it starts with a `/`: `url.path.startsWith("/")`.
We could also have a `hasAbsolutePath` property to do this check, like Dart and furl.

Other languages do different things.
.NET keeps the trailing slashes in the segments (weird):
`http://foo.com/bar/baz` => `["/", "bar/", "baz/"]

Swift adds the leading `/` but removes it from the middle segments:
`http://foo.com/bar/baz` => ["/", "bar", "baz"]

In Rust `path_segments` returns an optional, and is only set for paths with a root, so the ambiguity doesn't exist.
Same for Java: `getPath()` is null for an opaque URI.

spices/SPICE-0033-url-library.adoc
stackoverflow: I made these use `Mapping`/`Listing` to be in sync with `buildQuery` and `parseQuery`. So one can do `buildQuery(url.queryParameters)`.

I think we should keep these 3 in sync for ergonomics, but I think it's fine to move all of them to `Map`/`List`.

spices/SPICE-0033-url-library.adoc
stackoverflow: Added normalize.

spices/SPICE-0033-url-library.adoc
stackoverflow: > Any reason to mark these local?

I marked them as `local` because they are only used for validation. I don't think there's a solid use case for exposing them.

Moved the functions inside `Url`.

spices/SPICE-0033-url-library.adoc
stackoverflow: Indeed. Missed that on my copy-paste...

spices/SPICE-0033-url-library.adoc
bioball: This current solution, .NET's, and Swift's all feel like it introduces edge cases for no particular reason.

I don't mind Rust's solution, though.

Java's `getPath()` for opaque URIs is strictly incorrect according to RFC-3986 (`foo:bar` has path `bar`).

spices/SPICE-0033-url-library.adoc
bioball: I think this is another reason that we should have different fields for percent-encoded values and their raw counterparts.

If you're constructing one component-wise, you really want to know if you should be treating the value as already encoded or not. For example, Java assumes that the component-wise constructor is _NOT_ percent-encoded.

```java
new URI("file", "", "/foo%20bar.txt", null).toString()
// gives you "file:///foo%2520bar"

new URI("file", "", "/foo bar.txt", null).toString()
// gives you "file:///foo%20bar"
```

I think it would be bad for these two to have the same semantic meaning:

```pkl
new net.Url { path = "/foo%20bar" }
new net.Url { path = "/foo bar" }
```

spices/SPICE-0033-url-library.adoc
stackoverflow: We can solve the edge case in the current solution by adding a `hasAbsolutePath` like Dart.

Rust's solution is not bad, but it's more in line with WHATWG (which is the spec the library follows). RFC 3986 explicitly accepts rootless paths:
```
path-absolute = "/" [ segment-nz *( "/" segment ) ]
path-rootless = segment-nz *( "/" segment )
```
So, returning `null` is strange.
It also doesn't solve the problem: if you want the decoded segments for a path regardless of root, you'll have to split and decode the path by hand, which is way more annoying that check a boolean.

spices/SPICE-0033-url-library.adoc
stackoverflow: Removed the `local` modifier.

spices/SPICE-0033-url-library.adoc
stackoverflow: Now the components are always properly encoded. There are new `hidden fixed raw*` properties for the decoded version. Amendment requires proper encoding so `new net.Url { path = "/foo bar" }` will error.

